PR #2, opened 2021-01-25
feat: add option to configure width and border style

elianiva:
This PR adds 2 new options to set the max width of hover window and its border style.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,9 @@ local saga = require 'lspsaga'
 -- finder_definition_icon = '  ',
 -- finder_reference_icon = '  ',
 -- definition_preview_icon = '  '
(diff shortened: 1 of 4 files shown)


glepnir:
LGTM:) Thanks.

glepnir:
Bug with max_hover_widht it will hung neovim.

elianiva:
Ah, my bad. I didn't test if it has `0` value.

elianiva:
Ok I have a fix for that, turns out it `_make_floating_popup_size` is using `max_width` that we provided. The solution is just renaming `max_width` in `handler.lua` to `max_hover_width`. Do you want me to make a PR for this?

PR #6, opened 2021-01-25
fix: max_hover_width causing neovim to freeze

elianiva:
follow up of #2, this won't freeze neovim


glepnir:
Thanks I will test it later.

PR #23, opened 2021-01-26
chore: remove prints

rcarriga:
Just removing some spammy prints. Guessing they were leftover from some debugging.


glepnir:
Aha. Thanks for PR. but there are some new change . I had remove it.

rcarriga:
Perfect no worries!

Issue #24, opened 2021-01-26
Use project relative paths in finder window


weilbith:
Hey,
so great to see your plugin becoming better and better! :rocket: 

I recognized that in the window with all the references etc using the `lsp_finder` function, all paths are relative to the users home directory. At least for me this is the case. And for me this sometimes end up with really long paths. Which make the window huge and "bad" to read. What would you think about making them relative to the projects root? I'm not sure if it is possible to get the root as detected by the LSP server. But probably it would be fine to set them relative to NeoVims current directory as well. Not that powerful, but better then relative to `$HOME`. Wouldn't it? :thinking:

glepnir:
Now I had used the lsp root dir or home dir . but some file path still too long.
https://github.com/glepnir/lspsaga.nvim/blob/b01607b22b02c80c004bc2ba8b94263b0a453a04/lua/lspsaga/provider.lua#L58-L63

weilbith:
Hmm. I just did an update and are locally now on b01607b which includes your references changes. But I still get the home relative path. When I open `:LspInfo`, the root of the language server is correct. For all of them (I use `efm` as general purpose language server for linting and formatting next to the "regular" servers). Any idea how I could debug this?

glepnir:
can you post the min code and which server? now I am work with remake auto preview window.

PR #29, opened 2021-01-26
Fix truncation of paths for finder provider

weilbith:
Fixes: #24

See the commit message for the details.

### DIFF
--- a/lua/lspsaga/provider.lua
+++ b/lua/lspsaga/provider.lua
@@ -55,10 +55,10 @@ local create_finder_contents =function(result,method_type)
       local short_name
 
       -- reduce filename length by root_dir or home dir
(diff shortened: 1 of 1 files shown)


weilbith:
Btw: @glepnir do you use Windows?

glepnir:
I don't use windows.

weilbith:
> I don't use windows. 

Ah okay. Was just a wild guess why you might not observed this issue yourself. Because of the backslashes in Windows paths. :shrug:

Thanks for accepting the PR. :hugging:

glepnir:
There has a problem that if filepath does not match root_dir or home dir it will got nil so i add a patch for this

weilbith:
Yes makes sense to add this case. :+1:

PR #36, opened 2021-01-28
Fix typo in README for LspSagaRename command name

weilbith:
Nitpick...

Btw: I can't find it in the docs, but I think if you use the `<cmd>` prefix (which has many advantages next performance) you don't need to silence the commands. You use silence for with and without the `<cmd>`. Do you really need it? :thinking:  And is there a specific reason you mixed both variants? Just asking... :speak_no_evil:


glepnir:
Everyone has different habits, for me the command is simpler

Issue #30, opened 2021-01-26
[Feature] Current line diagnostic and signature help


kristijanhusak:
Hi,

what do you think about adding current line diagnostic and signature help? Currently we have these in neovim:

Diagnostic:
```
vim.lsp.diagnostic.show_line_diagnostics()
```

Signature:
```
vim.lsp.buf.signature_help()
```

And I'm using it like this:
```lua
  vim.cmd [[autocmd CursorHold * lua vim.lsp.diagnostic.show_line_diagnostics()]]
  vim.cmd [[autocmd CursorHoldI * silent! lua vim.lsp.buf.signature_help()]]
```

But it would be nice to have this nice styling to it.

Thanks!

glepnir:
signature help support now
![signature_help](https://user-images.githubusercontent.com/41671631/105969051-c7fb7700-60c2-11eb-9c79-aef3e01d88b1.gif)

kristijanhusak:
Signature works great, thanks. Only current line diagnostics are left.

nghialm269:
@glepnir The signature help popup currently use the current buffer's filetype to highlight, which makes some random words get highlighted:
![image](https://user-images.githubusercontent.com/13317164/106230741-c5179800-6222-11eb-8d21-a7304af695bc.png)

Is it possible to use the current buffer's filetype to highlight the label, and markdown to highlight the documentation?

glepnir:
There are some problems with the render doc in neovim upstream.When I implement other features, I will  try to make it better

glepnir:
@kristijanhusak Is the display line diagnosis really useful? I haven't used it much. I want to use a floating window to display all the diagnostic information of the buffer and add some previews. Of course, you can jump from this window to the location of the diagnosis, which is more useful than the line diagnostics?

kristijanhusak:
@glepnir yeah, it's useful for me. I disable virtual text diagnostic because it creates too much noise for me. I have a CursorHold event that shows current line diagnostic (see first comment), and i know where errors are by sign column. If i want to see all, i just open up quickfix window.

glepnir:
support

kristijanhusak:
@glepnir I gave current line diagnostic a try, it works, but it's a bit different when you jump, and when you preview. For example, this is what I get when I use current line diagnostic:

![current-line](https://i.imgur.com/sXTQXaw.png)

And this is when i jump to it with "next":

[...]

PR #42, opened 2021-01-29
Use highlight groups to adjust color appearance

### LINKED ISSUE #41: Use custom highlight group for selection in finder
You recently added the ability to change colors to highlight the current active entry in the finder. This is configured by two color values. I think it would be cool if you would add a new custom highlight group and then link the cursor line to this group. This still allows the user to easily change the highlight. But instead of defining raw color values you could also create a highlight link. So for example would I link it to the `Search` highlight group of my theme. But now I need to copy paste the colors. And I can't even apply "effects" like bold text. And now if I change my theme, I must also change the colors here again...
Would be a cool addition.

---

Btw: your recent improvements on the finder window are great! :muscle:

weilbith:
Fixes #41

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,8 +32,6 @@ local saga = require 'lspsaga'
 -- border_style = 1
 -- max_hover_width = 0 (automatically adjust to the width of current symbol)
 -- rename_prompt_prefix = '➤',
--- selected_fg = '#1c1f24',
--- selected_bg = '#b3deef'
 
 saga.init_lsp_saga {
   your custom option here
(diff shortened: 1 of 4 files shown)


glepnir:
Then I think we do not need these code right?
https://github.com/glepnir/lspsaga.nvim/blob/main/lua/lspsaga/provider.lua#L360-L365

weilbith:
> Then I think we do not need these code right?
> https://github.com/glepnir/lspsaga.nvim/blob/main/lua/lspsaga/provider.lua#L360-L365

Oh, I haven't seen this. What's its exact purpose? I don't understand it completely...

glepnir:
I got a problem that if we quit the `lsp finder` window. the highlight of `CursorLine` need reset .

weilbith:
Ah yeah. This should be still an issue, as we use a regular group. If we would do the highlighting ourselves this would be fine. But for changing the cursorline we still need this reset logic.
So for the moment it still works as I see it. We could think about alternating the solution to get rid of the "get back to old highlight" logic.

weilbith:
Why did you reverted it? It should just work. Or what is the issue?

glepnir:
when i quit the `finder` window ..the `CursorLine` use `LspSagaSelction` bg color ..so I revert it. didi you test it in your local ?

weilbith:
Really? For me it does not. I mean in fact it is the same as before. Your reset logic should be still there. I basically just changed the definition part...

weilbith:
I tried it again. And it works. So do you have an explanation why your former reset logic does not work anymore?

glepnir:
hmm why does not work for me . I run `git pull` then does not work..so revert it sorry. can you create a pr again I will check what happen . current I am fix the issue of diagnostic.

weilbith:
> hmm why does not work for me . I run git pull then does not work..so revert it sorry. 

No problem. Correct reaction.

> can you create a pr again I will check what happen .

Sure. :100: 

> current I am fix the issue of diagnostic.

What is it about? I don't see an open issue regarding this.

glepnir:
check here #30

PR #47, opened 2021-01-30
remove overwrite diagnostic handler

glepnir:
@weilbith  Hi  this pr remove over write default `handlers['textDocument/publishDiagnostics']` , but we also support in built in 
everyone can change the `diagnostic_opts` with their own options. what do you think?


weilbith:
Hey, thanks for asking me. :hugs: 

At first I would like to ask you what is your goal with the diagnostic handler in this plugin? Your plugin has great additions to the default LSP functions/features, especially for the visual part. All of them are extensions and do not interfere with the default NeoVim interface (after you have removed the hover handler). This is great, making your plugin minimal invasive to the users setup, but still powerful.
Now there is the diagnostics handler of this plugin. In fact it doesn't add anything new to NeoVim. The only thing it can do is providing cool defaults for "lazy" users so it just looks good. Therefore I think the already existing boolean option in the configuration is good enough. Use the plugin's defaults or do it yourself. But now it looks like you are adding an interface to the interface that NeoVim already provides. Why should users set the diagnostic handler options via your plugin if they could do it themselves directly and independently of your plugin? It is more work for you to maintain the interface you don't control. And for the user you don't add any new feature they don't already have by plain NeoVim.
So I would just stick with what you do good. And for this specific case it is providing sane default configurations, if the users does not want to do it himself. But if he likes to (as I do for example) when they can just disable it and still profit from all the other great features.

Was that somehow helpful? I think you are doing a great job. :+1:

glepnir:
Maybe I need remove overwrite diagnostic handler.

Issue #48, opened 2021-01-30
Large floating windows are not truncated vertically


akinsho:
Thanks for the great work on this plugin 🚀 

Ive noticed whilst trying it out that when I hover over something with a lot of information the windows don't seem to have safe default limits and in the case of the vertical size, the height isnt truncated if there is too much content like in the example below, it overlaps with my tabline and statusline in this case

![image](https://user-images.githubusercontent.com/22454918/106366633-17e17480-6335-11eb-80de-285a95d29fdc.png)

glepnir:
yep we can runcated vertically but need a smart height value to do that.

akinsho:
@glepnir I think
```lua
-- As an example I think 80% of the "safe area" is a good default
local max_size = (vim.o.rows - vim.o.cmd_height - (statusline_size_is_one) - (tabline_height_is_one)) * 0.8
```

akinsho:
The 80% is arbitrary but you get the idea aka it should not eat into other vim UI like tabline and statusline and tbh even then I dont think it's useful for it to take up *all* of the "safe" vertical height I think it should still only be a proportion of the vertical axis with the that space

glepnir:
and then we need support scroll in doc float window.

akinsho:
Tbh I think this is needed anyway since even with the current behaviour as you can see with my screenshot it still isn't guaranteed you will be able to fit all the content into the window

glepnir:
yep . now I had rewrite the width in my local .

glepnir:
how about this
![image](https://user-images.githubusercontent.com/41671631/106383446-f9e14600-6400-11eb-8e2d-3aca696fb32e.png)

PR #52, opened 2021-01-31
enhance window behavior

### LINKED ISSUE #50: feat: enhance hover doc format
Hi there! Is it possible to make the first row on the hover doc to be multiple lines? eg.

image

I expect the first separator line to get placed at the line where that is marked with a red line. This is causing some issue when it doesn't have enough space until the `{` symbol so the highlight got messed up. Here's an example

image

Thanks in advance!

glepnir:
Enhance the hover doc window behavior. try close #50  #48

@elianiva  @akinsho Hi can you test this patch


akinsho:
Will try in a few hours not by my computer at the moment

glepnir:
Okay,not a complete patch. need support scroll in doc buffer.

glepnir:
use `lua require('lspsaga.hover').scroll_in_hover()` to scroll hover doc

elianiva:
I believe this is still a WIP because the window stays there when I move my cursor :laughing:

image

Here's the error message.

```
Error executing vim.schedule lua callback: ...site/pack/packer/opt/lspsaga.nvim/lua/lspsaga/window.lua:321: Cannot convert given lua table
```

glepnir:
@elianiva  can you test now?

elianiva:
Great! It works now.

akinsho:
@glepnir I still haven't tried it yet and it doesn't have to be part of this change but it looks like the scrolling only goes in one direction, why not take an argument like -1 or 1 so you can map to different keys and go up and down

glepnir:
I forget scroll up support it later

akinsho:
Awesome just tried this out @glepnir and it works nicely 🚀 great work.

When you get to doing scrolling (no rush) just a small recommendation would be to do it similarly to how `coc` handles the mapping e.g.
```lua
  map("n", "<C-f>", [[coc#float#has_scroll() ? coc#float#scroll(1) : "\<C-f>"]], nowait)
```
basically it provides a utility function to see if the floating window for the plugin is open then allows you to map it in a way that you can use normal scroll commands but if the hover window is open then it will scroll that instead

akinsho:
Sorry one last point, also just realised that this doesn't apply to the finder window
image

elianiva:
Probably not related but is there any way to attach an autocmd to lsp_finder? I want to attach an autocmd that hides my cursor but the filetype is `plaintext` and the buftype is `nofile` which isn't specific to lsp finder.

glepnir:
@akinsho  @elianiva  new change

1. custom finder action keymap 
```
finder_action_keys = {
   open = 'o', vsplit = 's',split = 'i',quit = 'q'
},
```
2. smart height width of finder and hover doc so I think can remove `max_hover_doc_width`
3. support scoll hover doc not jump into it
```lua
lua require('lspsaga.hover').has_saga_hover() --true mean has hover doc --false not
lua require('lspsaga.hover').scroll_in_hover(1 or -1)  -- 1 mean scroll down -1 mean scroll up
```

glepnir:
when test ok I will merge.

akinsho:
🤔 just tested it out and the finder window is still really long not sure if I'm missing something

glepnir:
now should be  works.

akinsho:
Awesome that works now 👍🏿

glepnir:
@akinsho  If no other problems I will merge this pr.

akinsho:
Sounds good to me, this resolves my height issues 💯

elianiva:
ditto, everything works well.

akinsho:
For the future @glepnir I just noticed the signature window behaves the old way, maybe that could needs updating as well
image

glepnir:
support it tomorrow

glepnir:
@akinsho  done.

glepnir:
smart scroll hover doc, if window has hover doc it will scroll hover , if not it will do <C-f>
```lua
-- scroll down hover doc
nnoremap <silent> <C-f> <cmd>lua require('lspsaga.hover').smart_scroll_hover(1)<CR>
-- scroll up hover doc
nnoremap <silent> <C-b> <cmd>lua require('lspsaga.hover').smart_scroll_hover(-1)<CR>
```

akinsho:
Just tested this out @glepnir I still see

image

It looks like it's moved but the width is still really small. You can reproduce by typing `print(<cursor>`.

glepnir:
`echo winwidth(0)` output

glepnir:
should be works now.

akinsho:
@glepnir yep that's fixed it thanks 👍🏿

PR #53, opened 2021-02-02
async lsp finder

### LINKED ISSUE #33: [Discuss] Async render lsp finder
For some language servers with very slow performance, such as pyls, lspfinder uses buf_request_sync, which awaits the response and renders the window. This is a blocking behavior. If the language server responds very slowly, it means waiting for it, which is very bad.

1. use `vim.loop.new_async`
2. use `buf_request` need rewrite render window.

glepnir:
async lsp finder enhance for some language servers with very slow performance, close #33

### DIFF
--- a/lua/lspsaga/provider.lua
+++ b/lua/lspsaga/provider.lua
@@ -23,7 +23,7 @@ local send_request = function(timeout)
   for i,response in ipairs(responses) do
     if type(response) == "table" then
       for _,res in pairs(response) do
-        if res.result and next(res.result) ~= nil then
+        if res.result then
           coroutine.yield(res.result,i)
         end
       end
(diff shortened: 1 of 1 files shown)


weilbith:
As this is kinda hard to try out, but I'm curious and lazy. :grimacing: 
Would this meant the window gets renders instantly and then the results get inserted when they arrive? :thinking:

glepnir:
No, getting the results and window rendering are executed sequentially. I wrapped them into a new thread, which means that neither the request result nor the window rendering will block the current behavior.

weilbith:
Hmm okay. And how does that help? :thinking: 
I mean when I call any function of this plugin I'm await a result. It is not like I'm triggering it and continue doing other stuff. Don't you?
But anyway, async features are mostly very cool. And despite it might not be too helpful it is nice to have. :+1:

glepnir:
Yes, when we use finder. We hope to wait for the result, but for some server pyls and so on, they may take 2s-5s to get a response, which will bring a very bad experience. Using a new thread will reduce the response time.

> Would this meant the window gets renders instantly and then the results get inserted when they arrive?


It’s not difficult to do, but the result will look strange, because we got the window but haven’t got a response yet,Insert contents into window when the result is obtained. Also for servers with poor performance, we may see the result inserted 2s after the window appears.

weilbith:
> Yes, when we use finder. We hope to wait for the result, but for some server pyls and so on, they may take 2s-5s to get a response, which will bring a very bad experience. Using a new thread will reduce the response time.

Yes but we would wait anyway right? So it is not like after one second I still have no result, I continue to write. Because either I was in need of some information or I wanted to navigate. But anyway, cool feature.

> It’s not difficult to do, but the result will look strange, because we got the window but haven’t got a response yet,Insert contents into window when the result is obtained. Also for servers with poor performance, we may see the result inserted 2s after the window appears.

Well we could do it like in any web-application: put a progress indicator. This could a pretty ugly text or advanced stuff like a spinner. I know for example that [wilder.nvim](https://github.com/gelguy/wilder.nvim) has a spinner. And I could imagine that this would make this plugin look sick. :rainbow:

glepnir:
The purpose of using a new thread is only to reduce some of the time to get a response, although the improvement is not big, because the performance is all on the language server side, not for users to continue other things, when users use finder, they are all to wait for the results. So I think this shouldn’t be a problem

Using a spin may be good, it looks good, I will try it when I have time, interesting ideas

PR #60, opened 2021-02-03
Fix server_filetype_map.

### LINKED ISSUE #57: Error using `lsp_finder()` for someone not using `nvim-lspconfig` but still has a valid LSP connection.
Give the following init file:

```
local cmd = vim.cmd

cmd [[packadd packer.nvim]]
require('packer').startup(function(use)
  use {'wbthomason/packer.nvim', opt = true}

  use {'glepnir/lspsaga.nvim'}
  use 'scalameta/nvim-metals'

end)

local saga = require 'lspsaga'
saga.init_lsp_saga()

vim.o.shortmess = string.gsub(vim.o.shortmess, 'F', '') .. 'c'

local function map(mode, lhs, rhs, opts)
  local options = {noremap = true, silent = true}
  if opts then
    options = vim.tbl_extend('force', options, opts)
  end
  vim.api.nvim_set_keymap(mode, lhs, rhs, options)
end

map('n', 'gr', '<cmd>lua require"lspsaga.provider".lsp_finder()<CR>')
cmd [[augroup lsp]]
cmd [[autocmd!]]
cmd [[autocmd FileType scala,sbt lua require("metals").initialize_or_attach({})]]
cmd [[augroup end]]
```
When I attempt to use `lsp_finder()` I get the following error message:

```
Error executing vim.schedule lua callback: ...site/pack/packer/start/lspsaga.nvim/lua/lspsaga/libs.lua:21: bad argument #1 to 'pairs' (table
expected, got nil)
```

I'm still digging into finding out what the issue is, and I can give you detailed instructions on how to use `nvim-metals` if you want, but I'm curious if it has to do with the fact that I don't use `nvim-lspconfig` at all to manage the connect with metals, but rather my own plugin handles that all. Could that potentially be the issue here?

ckipp01:
This closes #57

This makes a few changes to the current `server_filetype_map`
functionality.

1. The logic used to be that you checked for `client.root_dir` first,
   and if that wasn't found, then you would look into the map. Instead,
   you should look to see if `config.filetypes` is defined first, if
   not, then you look into the map.
2. Instead of using `config.cmd` in the mapping, use `config.name`.
   Mainly because you can't always rely on `config.cmd` being a single
   executable. In the case of `nvim-metals` is a full path, in
   `nvim-jdtls` is a path to a script. So if you use name, the mapping
   becomes much simpler.
3. Swap around the value and key in the mapping so that a single server
   name can point to multiple filetypes. This is inline with how
   `config.filetypes` is, so probably best to keep the same. This is
   also important for Metals as we want to target both Scala and sbt
   files.

So now you would use this like so:

```lua
require('lspsaga').init_lsp_saga({
  server_filetype_map = {metals = {'sbt', 'scala'}}
})
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # lspsaga.nvim
 
-A light-weight lsp plugin based on neovim built-in lsp with highly performance UI.
+A light-weight lsp plugin based on neovim built-in lsp with highly a performant UI.
 
 ## Install
 
@@ -40,9 +40,9 @@ local saga = require 'lspsaga'
 -- 1: thin border | 2: rounded border | 3: thick border
 -- border_style = 1
 -- rename_prompt_prefix = '➤',
--- if you don't use nvim-lspconfig you must pass your filetype and server
--- command into this table
--- like server_filetype_map.scala = 'metals'
+-- if you don't use nvim-lspconfig you must pass your server name and
+-- the related filetypes into this table
+-- like server_filetype_map = {metals = {'sbt', 'scala'}}
 -- server_filetype_map = {}
 
 saga.init_lsp_saga {
--- a/lua/lspsaga/libs.lua
+++ b/lua/lspsaga/libs.lua
@@ -93,22 +93,22 @@ function libs.get_lsp_root_dir()
   if not active then print(msg) return end
   local clients = vim.lsp.get_active_clients()
   for _,client in pairs(clients) do
-    if client.config.root_dir then
-      if client.config.filetypes then
-        if type(client.config.filetypes) == "table" then
-          if libs.has_value(client.config.filetypes,vim.bo.filetype) then
-            return client.config.root_dir
-          end
-        elseif type(client.config.filetypes) == "string" then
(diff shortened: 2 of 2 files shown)


jlesquembre:
I found this issue looking for documentation about how to combine lspsaga and jdtls, could you provide and example? Is just

```lua
require('lspsaga').init_lsp_saga({
  server_filetype_map = {jdtls = {'java'}}
})
```
?

ckipp01:
> I found this issue looking for documentation about how to combine lspsaga and jdtls, could you provide and example? Is just
> 
> ```lua
> require('lspsaga').init_lsp_saga({
>   server_filetype_map = {jdtls = {'java'}}
> })
> ```
> 
> ?

Very close, but you'll want the name that the server sets: https://github.com/mfussenegger/nvim-jdtls/blob/master/lua/jdtls/setup.lua#L67

Which looks to be `jdt.ls`.

jlesquembre:
Thanks @ckipp01

PR #75, opened 2021-02-06
Allow to overwrite all highlight groups in a colorscheme 

### LINKED ISSUE #63: Allow to overwrite all highlight groups in a colorscheme
Currently it is not possible to customize the colors of all highlight groups in a colorscheme. 

The problem is that `lspsaga.nvim` defines and overwrites all groups after calling `init_lsp_saga(opts)` with a call to https://github.com/glepnir/lspsaga.nvim/blob/main/lua/lspsaga/syntax.lua.

My current workaround is to copy the function found in `syntax.lua` and call my custom function after the call to `init_lsp_saga(opts)`

```lua
require("lspsaga").init_lsp_saga({
    use_saga_diagnostic_sign = false,
    rename_prompt_prefix = '>',
    max_diag_msg_width = 80,
})

-- color scheme can not be used as the lspsaga plugin would overwrite the colors again
-- this function gets called after the lspsaga plugin was loaded and defines the final colors
require("settings.lspsaga").overwrite_highlight()
```

It would be nice if all the highlight groups used by the plugin can be overwritten in a colorscheme instead of calling a custom function after the plugin was loaded.

This relates to:
- #41
- #42

sblumentritt:
This resolve #63.

Move highlight group definition to vimscript file
This change allows colorscheme provider to change the colors used by the 'lspsaga.nvim' plugin.

----

I did not add a description for the highlight groups as some names are self explaining and the purpose of some other highlight groups are unclear to me.

In my opinion this plugin has to much highlight groups. 😄


glepnir:
LGTM

PR #76, opened 2021-02-06
try to fix #67

### LINKED ISSUE #67: Odd floats when in narrow screen.
I'm not sure what's the most helpful information to give you to chase this down, but please let me know what I can provide to help. My workflow involves often splitting my terminal with tmux so that nvim is only on half my screen. When this is done often any bigger floats get slammed up against the side. The border actually looks fine, but it's like the text shifts over a bit too far. Here is an example of what I mean while in Lua.

Screenshot 2021-02-04 at 20 38 43

Here is another example in Scala using Metals when I have full screen:

Screenshot 2021-02-04 at 20 40 32

However, if I then split my screen and do the hover on the same symbol it looks like this:

Screenshot 2021-02-04 at 20 41 01

glepnir:
@ckipp01 can you test this PR. not sure can fix  and works in every where .but  I test it works for me.

### DIFF
--- a/lua/lspsaga/window.lua
+++ b/lua/lspsaga/window.lua
@@ -167,7 +167,8 @@ function M.create_float_contents(content_opts,border_option)
   end
 
   if opts.anchor == "NE" then
-    opts.col = border_option.col - 1
+    opts.anchor = "NW"
+    opts.col = -13
   elseif opts.anchor == "NW" then
     opts.col = border_option.col + 1
   else


ckipp01:
Just gave it a test drive and it looks like I can still reproduce:

Screenshot 2021-02-06 at 14 51 01

glepnir:
Is there any change when adjust -13 to -10 ?

ckipp01:
🤔 so with `-10` it's the exact same.

glepnir:
Need check more detail, -13 looks fine for me 
image

PR #82, opened 2021-02-08
feat: Expose hover handler to be used as a handler

tjdevries:
Hey @glepnir , I'd like to be able to do this:

```lua
vim.lsp.handlers["textDocument/hover"] = require('lspsaga.hover').handler
```

to be able to to make `lspsaga` just integrate into the rest of my builtin LSP stuff.

In general, I think it'd be cool to expose the handlers to be consumed a bit more independently if you don't mind.

Anyway, thanks, just checking out the plugin today :)

### DIFF
--- a/lua/lspsaga/hover.lua
+++ b/lua/lspsaga/hover.lua
@@ -17,7 +17,7 @@ local function focusable_float(unique_name, fn)
   end
 end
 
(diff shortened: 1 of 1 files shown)


glepnir:
Love you

tjdevries:
<3

PR #84, opened 2021-02-09
use codeaction quit key on rename

windwp:
I want to use <esc> for close rename . 
i add this mapping to reuse code action quit key

### DIFF
--- a/lua/lspsaga/rename.lua
+++ b/lua/lspsaga/rename.lua
@@ -57,6 +57,8 @@ local rename = function()
   api.nvim_buf_add_highlight(cb, saga_rename_prompt_prefix, 'LspSagaRenamePromptPrefix', 0, 0, #prompt_prefix)
   vim.cmd [[startinsert!]]
   api.nvim_win_set_var(0,unique_name,{cw,bw})
+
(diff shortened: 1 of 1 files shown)


glepnir:
There has a problem if the new name of func has a word `q` .the default quit key will be trigger. then quit the rename window.

windwp:
Now I only add esc to quit on  insert mode.If  i use a default action rename  on lsp i can esc to quit.

glepnir:
I think we can make it optional ? like` rename_keys = { quit = 'q',exec='<CR>'}`

windwp:
but you can't map q in insert mode. the only key can do that in insert mode is <esc>.

glepnir:
just example I mean config like code_action keys. that user can custom it.

windwp:
in insert or normal mode?

glepnir:
```lua
rename_keys = { quit = 'q' or table type {'q','<Esc>','<C-c>'}   ,exec = '<CR>'}  
```
quit is keys in insertmode . normal mode we can hardcored q  and esc ?

glepnir:
close by ad0ce56

PR #87, opened 2021-02-09
Windows lsp_finder fixes

HudsonMC16:
vim.loop.os_uname().sysname appears to return "Windows_NT", so the
in_windows check was changed to check for a substring instead of
straight equality. This was also migrated to a function to perform a
similar check in provi.ers.lua.

os.getenv("HOME") returns nil on Windows. os.getenv("USERPROFILE") is
appropriate on Windows.

vim.uri_to_fname() returns paths with lowercase drive letters on Windows
for some reason. Before the paths are shortened for the lsp_finder
results, a check for Windows was added, along with a substitution of the
drive letter for the appropriate uppercase version.

I'm no lua wizard, but I took a swing at it. I'm not sure if I caught all the Windows edge cases. I basically just worked through the hangups until I got lsp_finder working. Let me know if there is something else which should be looked into.

### DIFF
--- a/lua/lspsaga/libs.lua
+++ b/lua/lspsaga/libs.lua
@@ -1,10 +1,17 @@
 local api = vim.api
-local is_windows = vim.loop.os_uname().sysname == "Windows"
-local path_sep = is_windows and '\\' or '/'
 local libs = {}
 local server_filetype_map = require('lspsaga').config_values.server_filetype_map
 
+function libs.is_windows()
+  return vim.loop.os_uname().sysname:find("Windows", 1, true) and true 
+end
+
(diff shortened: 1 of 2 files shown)


glepnir:
also should be fixed in `libs.split_by_pathsep` function. I don't use windows.

HudsonMC16:
Ah, oops missed that one. Thanks. Should be fixed now. I'll continue to use the plugin and let you know if I run into any more issues on Windows.

glepnir:
Okay If you can confirm everything works fine in windows I will merge

HudsonMC16:
Yep, haven't experienced any issues yet. Appears to be working.

PR #91, opened 2021-02-11
use buf_request close #90

### LINKED ISSUE #90: Code Action not working when multiple LSP are attached
### Description
The default LSP handler works fine.

the lspsaga one and also telescope do not work for the same error.

**Expected Behavior**
I expect to see a list of actions

**Actual Behavior**
I get a error

### Details

So, this is what I can see in the LSP logs

[ ERROR ] 2021-02-10T16:35:32+0100 ] ...nt_nvimrAVbQs/usr/share/nvim/runtime/lua/vim/lsp/rpc.lua:455 ]	"rpc"	"/home/lucatrazzi/.cache/nvim/lspconfig/efm-langserver/efm-langserver"	"stderr"	'2021/02/10 16:35:32 jsonrpc2 handler: notification "$/cancelRequest" handling error: jsonrpc2: code -32601 message: method not supported: $/cancelRequest\n'

I wonder, if I have more than one LSP attached servers (efm and ts) and both have capability to give code_action, who is gonna answer to the json request? Is plugin gonna request both?

PS: I love how efm is giving capability and then saying "method not supported"
PS2: if I comment efm, it works.

glepnir:
@lucax88x  can you test this patch. it return to use `buf_request`

### DIFF
--- a/lua/lspsaga/codeaction.lua
+++ b/lua/lspsaga/codeaction.lua
@@ -23,17 +23,15 @@ function Action:action_callback(response)
     print("No code actions available")
     return
   end
+
   local contents = {}
   local title = config['code_action_icon'] .. 'CodeActions:'
   table.insert(contents,title)
-  self.actions = {}
 
-  for _,languageServerAnswer in pairs(response) do
-    for index,action in pairs(languageServerAnswer.result) do
-      local action_title = '['..index..']' ..' '.. action.title
-      table.insert(contents,action_title)
-      table.insert(self.actions,action)
-    end
+  self.actions = response
+  for index,action in pairs(response) do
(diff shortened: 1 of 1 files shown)


lucax88x:
Error executing vim.schedule lua callback: ...ack/packer/start/lspsaga.nvim/lua/lspsaga/codeaction.lua:32: bad argument #1 to 'pairs' (table expected, got nil)

glepnir:
fixed now

lucax88x:
Works, awesome. Thanks!

lucax88x:
I think that fix should be applied to all LSP calls.

glepnir:
yep. need to check other lsp calls

PR #95, opened 2021-02-12
Add method for diagnostics under the cursor

tapayne88:
First off, love the plugin!

I wanted something similar to my CoC setup which shows all LSP diagnostics under the cursor position. I don't _think_ there is a way to do that with the existing API so I've added a new function to expose it.

```lua
require("lspsaga.diagnostic").show_cursor_diagnostics()
```

What do you think?

### DIFF
--- a/lua/lspsaga/diagnostic.lua
+++ b/lua/lspsaga/diagnostic.lua
@@ -24,6 +24,36 @@ local function get_character(diagnostic_entry)
   return diagnostic_entry["range"]["start"]["character"]
 end
 
+local function get_line_last(diagnostic_entry)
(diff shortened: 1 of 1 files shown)


glepnir:
This mean show diagnostic msg for cursor word  right?

tapayne88:
Thanks for merging! Yeah, this should filter down all LSP errors for the current buffer to show just the errors for the current cursor position (likely the cursor word).

glepnir:
@tapayne88  Hmm I must say sorry to you. I had rewrite diagnostic with neovim upstream .so you need reimplement it in current diagnostic of lspsaga.

tapayne88:
Ah, no worries! I'll take a look 👍

PR #96, opened 2021-02-12
Add circular diagnostics jumps

kuznetsss:
Thank you for the plugin!
I suggest to jump to the first/last element of diagnostics messages if there are no messages above/below.


glepnir:
Thanks for Pr. LGTM

PR #118, opened 2021-02-21
fix signaturehelp

alex-popov-tech:
that pr fixes check for signature_help when `efm` is used ( previously if there `efm` first lsp it returned `false` which is not always correct )

### DIFF
--- a/lua/lspsaga/signaturehelp.lua
+++ b/lua/lspsaga/signaturehelp.lua
@@ -20,12 +20,8 @@ local function check_server_support_signaturehelp()
   local active,msg = libs.check_lsp_active()
   if not active then print(msg) return end
   local clients = vim.lsp.get_active_clients()
-  for _,client in ipairs(clients) do
-    if client.name == "efm" then
-      return false
-    end
-  end
-  return true
+  local isOnlyEfm = vim.tbl_count(clients) == 1 and clients[1].name == "efm"
+  return not isOnlyEfm
 end
 
 local function focusable_float(unique_name, fn)


glepnir:
Thanks for pr. make sence.

akinsho:
I'm not entirely sure how but this PR I think regressed the signature help `efm` fix for me 😭. I just tried reverting it and it seems to fix the regression. I tested/found this in a markdown file with only `efm` enabled. I tried debugging a little and I can't see what is different about the logic but it does seems to cause this bug again

alex-popov-tech:
oh oh, hi again @akinsho !
will try to reproduce today

alex-popov-tech:
hey @akinsho , can you provide some details?
image
when calling sig help it just shows msg and that's it

akinsho:
@alex-popov-tech I have a `CursorHoldI` autocommand which I use to call `lua require('lspsaga.signaturehelp').signature_help()`. In the `markdown` buffer where I'm getting the issue. I've actually just realised this issue is a little more complicated and maybe shouldn't be dealt with here. Basically If I go to a `lua` file for example then I enter a `markdown` file then `:LspInfo` shows

```

  Configured servers: html, efm, sumneko_lua, tsserver, jsonls, flow, vimls, rust_analyzer, dartls, gopls 
   
  2 client(s) attached to this buffer:  
   
  2 active client(s):  
   
    Client: efm (id 2) 
    	root:      /home/akin/.dotfiles 
    	filetypes: yaml, json, html, css, markdown, lua 
    	cmd:       efm-langserver 
   
    Client: sumneko_lua (id 3) 
    	root:      /home/akin/.dotfiles 
    	filetypes: lua 
    	cmd:       /home/akin/.local/share/nvim/lspinstall/lua-language-server/bin/Linux/lua-language-server -E /home/akin/.local/share/nvim/lspinstall/lua-language-server/main.lua 
   
  Clients that match the filetype markdown: 
     
    Config: efm 
    	cmd:               efm-langserver 
    	cmd is executable: True 
    	identified root:   /home/akin/.dotfiles 
    	custom handlers:    

```

Basically the lua server is somewhow attached to the markdown buffer and so the check doesn't work anymore and I get the error. I think there is a bug in `nvim-lspconfig` where buffers are being attached to incorrectly. The first fix by @glepnir I think worked since it didn't only check if just `efm` was attached but this change showed the underlying bug again. 

I think maybe there are 2 things here:
1. Why does `nvim-lspconfig` attach to lsps here? (maybe it's my error but I don't think so)
2. Even if `nvim-lspconfig` is buggy I still think calling signature help should fail silently instead of showing `E5108: Error executing lua .../packer/start/lspsaga.nvim/lua/lspsaga/signaturehelp.lua:137: attempt to index local 'result' (a number value)`

glepnir:
id is 2 so we can't use id 1 to check it.

glepnir:
revert. @akinsho  should be work now.

alex-popov-tech:
now it don't work for me....
i believe that actual issue in `id`, not in lspsaga check...
@akinsho can you do exact steps to reproduce? to achieve that state with two servers and no index `1`?
@glepnir meanwhile maybe I can try to just literate up to `10` and hope that all servers will be covered or something like that?

glepnir:
which error you get?

alex-popov-tech:
i have efm 1 and lua 2, so current check skips lua before even checking it, and sig help does not work at all

glepnir:
How about now?

alex-popov-tech:
better now thanks!
but the issue is still there right? you fixed just `md` filetype

glepnir:
yep quick patch. because I am improve the indent-guides.nvim now. will be enhance when I rewrite indent-guides. should be check yaml json etc that

alex-popov-tech:
the main issue - we should have func that tells if there is only `efm` there, so I come up with simple `if clients.length == 1 and clients[1] == efm` but it does not count case when second condition not met...

i still think that loop over all clients and check name, but the main point here is to insure we iterated over all of them...

so my proposition is to iterare up to some value like `10` assuming that there won't be more than 10 servers attached

alex-popov-tech:
i can create pr, if we agree on format :)

glepnir:
I think the quick way is like this,because these filetype does not need signaturehelp
```lua
local special_filetype  = {'yaml','markdown','json'}
if special_filetype[vim.bo.filetype] then
    return 
end
```

alex-popov-tech:
nice check, but that does not cover case when there is not a special type, and only `efm` attached, right?

glepnir:
yep this way will ignore some special type . then check the cliend id?

alex-popov-tech:
yes, that was initially built in this pr :)
and now we need somehow check if `efm` the only server attached

glepnir:
Yes, we can increase the priority of some checks， check filetypes return - > check it's only 1 client  is efm return -> latest is for client to check.

akinsho:
This is probably a distraction, but since `efm` doesn't support signature help rather than trying to special case it, I'm only guessing about what the issue is but why not similar to the earlier check you tried to implement @glepnir only forward trigger help requests to clients that support it like

```lua
function find_supported_client()
    local match
    for _, client in ipairs(vim.lsp.get_active_clients) then -- also need to filter against buffer
	  if client.resolved_capabilities.signature_help then
		 match = client
		 break
	  end
   return match
end

function signature_help()
	local supported_client = find_supported_client('signature_help') -- returns the first match or something
	if supported_client then
 		actually_trigger_signature_help(supported_client)
	end
end
```

This is just partly thought through pseudocode not sure it's a good idea to match just one but maybe that's all you can do 🤷🏿 anyway I just tried locally and `efm` has `resolved_capabilities.signature_help == false` so I think you can just depend on this value for all clients rather than special casing efm

alex-popov-tech:
good point, capabilities seems to be best option to detect lsp with sig help!

i will create pr and tag you guys to verify

glepnir:
First fix patch I used  `resolved_capabilities.` but don't know why .still got error.

akinsho:
@glepnir yeah I remember, maybe it was the right path but there was something missing. I can help test out a PR using that method, seems to consistently report support is `false` so must have been something else going on

alex-popov-tech:
please check out https://github.com/glepnir/lspsaga.nvim/pull/119
@glepnir @akinsho

PR #119, opened 2021-02-23
one more fix for signaturehelp

alex-popov-tech:


akinsho:
@alex-popov-tech that actually brings the bug back but I think the issue is that the loop breaks early if it meets any client that supports `signature_help` but I think the logic based on the way the file is laid out should be enforce a loop through all clients and only return `true` if *all* clients support signature help.

I think better still @glepnir not sure if this is possible but seems when we request signature help we can't specify the client we are requesting signature help for? I think it would be better if it's possible, could be an nvim limitation to somehow specify the client id(s) for  which you are requesting signature help

alex-popov-tech:
@akinsho why all? f.e. i have .ts file and `denols` for autocompletion/etc and `efm` for formatting/linting, so I can potentially can see sig help from deno, and `efm` is attached as well...
can you send exact output of `clients` I'm trying to iterate through? maybe I can fix the loop

akinsho:
@alex-popov-tech I think the issue is that `vim.lsp.buf_request(0, 'textDocument/signatureHelp)` that is being sent lower down in that file is not only sending the request for just the supported clients. It seems that the `is_supported` check will abort if `resolved_capabilities.signature_help` is `false` but that check will only work if it only hits `efm` first then returns false. If it has another client that supports signature help it seems it will go ahead and request it for all the clients maybe?

alex-popov-tech:
@akinsho now it iterates through all clients, and return `true` if at least one supports sig help, and `false` if no one supports that...so if you get to sending request with `efm` only, problem is in `check_server_support_signaturehelp` func

alex-popov-tech:
so if you have only `efm` it should send nothing, otherwise, there is a bug in my code

akinsho:
@alex-popov-tech so I get 
```
  Configured servers: html, efm, sumneko_lua, tsserver, jsonls, flow, vimls, rust_analyzer, dartls, gopls 
   
  1 client(s) attached to this buffer: efm 
   
    Client: efm (id 1) 
    	root:      /home/akin/.dotfiles 
    	filetypes: yaml, json, html, css, markdown, lua 
    	cmd:       efm-langserver 
   
  2 active client(s):  
   
    Client: efm (id 1) 
    	root:      /home/akin/.dotfiles 
    	filetypes: yaml, json, html, css, markdown, lua 
    	cmd:       efm-langserver 
   
    Client: sumneko_lua (id 2) 
    	root:      /home/akin/.dotfiles 
    	filetypes: lua 
    	cmd:       /home/akin/.local/share/nvim/lspinstall/lua-language-server/bin/Linux/lua-language-server -E /home/akin/.local/share/nvim/lspinstall/lua-language-server/main.lua 
   
  Clients that match the filetype markdown: 
     
    Config: efm 
    	cmd:               efm-langserver 
    	cmd is executable: True 
    	identified root:   /home/akin/.dotfiles 
    	custom handlers:    
```
Only one client is attached `efm` but `vim.lsp.get_active_clients` does not actually return only the *attached* clients I think what this needs is use `vim.lsp.buf_get_clients` then check if the client for that buffer match the check. Since atm it is checking both `lua` and `efm` even though only `efm` is actually active

Also the actual error is that on line `141` result comes back as a number in this case so I believe the check on that line should also check the `type(result) ~= "table" then return end`

```
E5108: Error executing lua .../packer/start/lspsaga.nvim/lua/lspsaga/signaturehelp.lua:137: attempt to index local 'result' (a number value)
```

alex-popov-tech:
@akinsho can you paste raw output of `vim.lsp.get_active_clients` and `vim.lsp.buf_get_clients` here? it returns the same for me

akinsho:
@alex-popov-tech will do in a bit, I'm supposed to be working atm 😅 but in short its pretty large so hard to paste since it doesn't fit on the screen. Checking the length of `vim.lsp.get_active_clients()` is `2` and `buf_get_clients` is `1`. Can you try opening nvim to a `lua` file then *after* opening a `markdown` file at this point if you run `LspInfo` you should see `2` active and `1` attached.

alex-popov-tech:
same thing :)
can you please share efm settings as well then? maybe I can reproduce your issue

akinsho:
Here's my `efm` config.
```lua
 efm = {
      init_options = {documentFormatting = true},
      filetypes = {"yaml", "json", "html", "css", "markdown", "lua"},
      settings = {
        -- add ".lua-format" to root if using lua-format
        rootMarkers = {".git/"},
        languages = {
          json = {prettier},
          html = {prettier},
          css = {prettier},
          yaml = {prettier},
          markdown = {prettier},
          lua = {
            {
              formatCommand = "luafmt --indent-count 2 --line-width 100 --stdin",
              formatStdin = true
            }
          }
        }
```

alex-popov-tech:
@akinsho thanks for the explanations! I've reproduced your issue, updated func to search in `buf_get` clients, can you please test once more? fix worked for me

akinsho:
@alex-popov-tech 💯  works like a charm 👍🏿 thanks for working on this

alex-popov-tech:
nice!
@glepnir mind merge that then?

glepnir:
Thanks for pr, LGTM merged.

Issue #144, opened 2021-03-17
LspSaga Hover window is closed too quickly


akinsho:
### Description

When using lspsaga's hover doc command to visual some info on a symbol, it closes itself very quickly which is quite frustrating because it becomes quite hard to read through the information it shows especially for languages that include a lot of docs

#### LSP Saga
![lsp_saga_hover_timeout](https://user-images.githubusercontent.com/22454918/111500334-c867d400-873b-11eb-8f3a-e6fe2b56e72b.gif)

#### Native LSP handler
![native_lsp_hover](https://user-images.githubusercontent.com/22454918/111500353-d158a580-873b-11eb-8674-3dbf2d7ef46c.gif)

**Expected Behavior**

It should remain open till the user performs an action

**Actual Behavior**

It closes after about 2 seconds

**Neovim Built in Behavior**

`vim.lsp.buf.hover()` stays open till the user moves as I've described

### Details

<details><summary>Reproduce</summary>

1. nvim -u test.vim the min config that only lspsaga
2. post your program language min code snippet that can reproduce your issue
3. 
</details>

<details><summary>Environment</summary>

- nvim --version output: 
- Operating system: 
- lspsaga commit: 

</details>

glepnir:
hmm did you use some autocmd for show line diagnostic or something else ?

akinsho:
Yeah I have
```lua
  require("as.autocommands").augroup(
    "LspSagaCursorCommands",
    {
      {
        events = {"CursorHold"},
        targets = {"*"},
        command = "lua require('lspsaga.diagnostic').show_line_diagnostics()"
      }
  }
```

Are they interacting? I would have thought that they were separate so one shouldn't close the other

glepnir:
I saw it in your dotfiles :) . a logic bug of saga .

glepnir:
@akinsho  sorry for bother you. can you confirm now works or not?

akinsho:
@glepnir that fixed it 👍🏿 thanks

akinsho:
It still closes if there is a diagnostic window on the line but thats fine I saw #138 and some other chat about unifying the windows which I think would be helpful in the future (if you decide to do it) but this is good for now

glepnir:
yep I need refactor the window api of lspsaga later

glepnir:
now should be works  :)

akinsho:
[...]

PR #147, opened 2021-03-18
Fix diagnostics conflicting with hover doc

tapayne88:
If I try render the hover window with `render_hover_doc`, `show_cursor_diagnostics` immediately dismisses the hover doc and render the diagnostics again (happens with `show_line_diagnostics` too). I _think_ this is due to `CursorHold` triggering after `render_hover_doc` which I have bound to show cursor diagnostics
```lua
vim.api.nvim_command('autocmd CursorHold <buffer> lua require("lspsaga.diagnostic").show_cursor_diagnostics()')
```

I've added a check to see if hover doc is present before rendering the diagnostic and provided a way to override it if the function is bound to something else.

### DIFF
--- a/lua/lspsaga/command.lua
+++ b/lua/lspsaga/command.lua
@@ -12,6 +12,7 @@ local subcommands = {
   preview_definition = provider.preview_definition,
   rename = lsprename.rename,
   hover_doc = lsphover.render_hover_doc,
+  show_cursor_diagnostics = diagnostic.show_cursor_diagnostics,
   show_line_diagnostics = diagnostic.show_line_diagnostics,
   diagnostic_jump_next = diagnostic.lsp_jump_diagnostic_next,
   diagnostic_jump_prev = diagnostic.lsp_jump_diagnostic_prev,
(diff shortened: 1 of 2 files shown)


tapayne88:
Think this is somewhat related to #144

glepnir:
can you test now ? I think should be work now.

tapayne88:
Yeah, it was still occurring after commit 7984e5a

glepnir:
hmm why it works for me ...

tapayne88:
Huh, that's odd 🤔 

This is what I'm seeing

https://user-images.githubusercontent.com/991180/111624038-14ba1f00-87e3-11eb-8c77-63bb015d6526.mp4

I guess these are the relevant bits from my dotfiles ([link](https://github.com/tapayne88/dotfiles/blob/master/dot_config/nvim/lua/_lsp.lua#L78))
```lua
vim.api.nvim_command('autocmd CursorHold <buffer> lua require("lspsaga.diagnostic").show_line_diagnostics()')
vim.api.nvim_buf_set_keymap(bufnr, 'n', 'K', '<cmd>lua require("lspsaga.hover").render_hover_doc()<CR>', opts)
```

glepnir:
I reproduce . so I merged 👍   Thanks for pr

PR #149, opened 2021-03-18
Apply severity sorting to all diagnostics before rendering

tapayne88:
I don't _think_ the [`opts.severity_sort`](https://github.com/glepnir/lspsaga.nvim/blob/main/lua/lspsaga/diagnostic.lua#L69) was doing anything so I've made it do what I think it should - happy to change it though


glepnir:
LGTM

PR #180, opened 2021-04-18
fix: not overwrite NormalFloat

tamago324:
Fixed to not overwrite `NormalFloat` if it has already been defined.

### DIFF
--- a/plugin/lspsaga.vim
+++ b/plugin/lspsaga.vim
@@ -19,7 +19,7 @@ highlight default LspSagaFinderSelection guifg=#89d957 guibg=NONE gui=bold
 highlight default LspFloatWinBorder guifg=black guibg=NONE
 highlight default LspSagaBorderTitle guifg=orange guibg=NONE gui=bold
 
-hi! link  NormalFloat Normal
+hi link  NormalFloat Normal
 highlight default link TargetWord Error
 highlight default link ReferencesCount Title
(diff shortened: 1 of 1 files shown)


tamago324:
Thanks!

akinsho:
Despite this change I still find that my `NormalFloat` is being overriden by Lsp saga. Can this addition of highlighting be done inside an autocommand or `ftplugin` so it only affects lsp saga windows. Currently it has changed the color of everything else I use which is quite frustrating

windwp:
I has the same problem too I have different color  NormalFloat and Normal so better don't link it.

tamago324:
As per the [comments in this issue](https://github.com/glepnir/lspsaga.nvim/issues/183#issue-862235856), it might be better to create a highlight other than `NormalFloat`.

I'm sorry if I'm wrong, but I didn't think `ftplugin` could handle this problem.
This is because I don't think the filetype of the floating window used by lspsaga can be fixed.

https://github.com/glepnir/lspsaga.nvim/blob/838bbf8141c0f37bcd35e1565831f1d1617db08e/lua/lspsaga/window.lua#L165

Therefore, it may be better to create a new `LspSagaNormalFloat` and set it to `Normal`.

akinsho:
I'm in favour of whatever works 👍🏿 yeah I guess the filetype of some windows will be that of the language inside it.

PR #185, opened 2021-04-20
fix(window): fiw renamed _trim method

martini97:
It seems that the `vim.lsp.util._trim` was renamed to `vim.lsp.util._trim_and_pad`, this was causing this error when trying to open the floaterm:
```
E5108: Error executing lua ...te/pack/packer/start/lspsaga.nvim/lua/lspsaga/window.lua:149: attempt to call field '_trim' (a nil value)
```


mjlbach:
I am the one who renamed the function, and it was renamed from _trim_and_pad to _trim (the opposite). You need to update your nightly.

PR #187, opened 2021-04-21
feat: add hi LspFloatWinNormal (close #183)

### LINKED ISSUE #183: Custom highlight group to override NormalFloat
In https://github.com/glepnir/lspsaga.nvim/commit/2752af4f3ccf9728dfc08a0079aaadc87ed066c4 all popup windows use `NormalFloat` instead of `Normal`. I have `NormalFloat` set to have a slightly different background color to `Normal`, but like having lspsaga's windows to be the same as the main background. Instead of using `NormalFloat` directly could it be changed to something like `LspSagaNormalFloat` that by default links to `NormalFloat`?

tamago324:
Added a highlight to be used for the floating window `Normal` used in lspsaga.

I named it `LspFloatWinNormal` to match the highlight name of `LspFloatWinBorder`.
If you don't like it, please let me know. I will fix it.

Refer #180, #183

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,6 +215,7 @@ The available highlight groups are:
 | Group Name               | Description                                                      |
 | :----------------------- | :----------------------------------------------------------------|
 | `LspSagaFinderSelection` | Currently active entry in the finder window that gets previewed. |
+| `LspFloatWinNormal` | |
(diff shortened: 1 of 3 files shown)


tamago324:
Thank you!

PR #194, opened 2021-05-07
Include line diagnostics without a range in cursor diagnostics

tapayne88:
I've found with some LSP servers they return diagnostics without a range. Sometimes the start and end char are `0` and in other cases they are the same value.

I have these reports in the gutter so I can see the line has an error however I have no way to view it (without using `show_line_diagnostics`).

I'm proposing to just include single line diagnostics with a `0` character range in all cursor diagnostics for that line.


glepnir:
can you update with the new diagnostic api in neovim 0.7?

glepnir:
@tapayne88 can you post a min code that can reproduce this error

tapayne88:
With this minimal config you should see the above error. You might need to open neovim with it (`nvim -u /path/to/minimal.lua`) and wait for the install to happen, then close and reopen.
```lua
-- To make use of this file run the following
-- nvim -u ~/.config/nvim/minimal.lua ...
vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvim/site]]

local package_root = '/tmp/nvim/site/pack'
local install_path = package_root .. '/packer/start/packer.nvim'
local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      {
        'williamboman/nvim-lsp-installer',
        requires = { 'neovim/nvim-lspconfig' },
        config = function()
          require('nvim-lsp-installer').setup {
            ensure_installed = { 'sumneko_lua@v2.5.6' },
            automatic_installation = true,
            install_root_dir = '/tmp/nvim/lsp_servers',
          }
          local lspconfig = require 'lspconfig'
          lspconfig.sumneko_lua.setup {}
        end,
      },
      {
        'tapayne88/lspsaga.nvim',
        branch = 'line-diags-without-ranges',
        config = function()
          require('lspsaga').init_lsp_saga()
        end,
      },
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. '/plugin/packer_compiled.lua',
      display = { non_interactive = true },
    },
  }
end

_G.load_config = function()
  require('nvim-lsp-installer').setup {}
end

if vim.fn.isdirectory(install_path) == 0 then
  print 'Installing packer packages...'
  vim.fn.system {
    'git',
    'clone',
    '--depth=1',
    'https://github.com/wbthomason/packer.nvim',
    install_path,
  }
end

load_plugins()
require('packer').sync()
vim.cmd [[autocmd User PackerComplete ++once echo "Ready!" | lua load_config()]]
```

If you open the `diagnostic.lua` file, wait for the language server to start and run `:Lspsaga show_cursor_diagnostics` on line 2 you should see the error.

[asciicast](https://asciinema.org/a/MHG7d76AykwFDRD35fqvkFahv)

Hope this helps!

glepnir:
yup

PR #235, opened 2021-08-09
Fix a mismatch of lightbulb highlight group name

chmnchiang:
Change the texthl argument for the lightbulb sign from `LspSagaLightBulbSign` to `LspSagaLightBub` so that it match the default highlight group definition in `plugin/lspsaga.vim`.


tami5:
Thanks @chmnchiang , merged temporary in https://github.com/tami5/lspsaga.nvim

glepnir:
@chmnchiang Add to Readme :)

PR #257, opened 2021-10-23
fix empty cursor_line_bg

kiddos:
I'm using a colorscheme that does not have cursor line bg. I will get error:
```
Vim(highlight):E417: missing argument: guibg=
```
when I close the finder window


glepnir:
seems like need rewrite ;)

kiddos:
oh, I already have this in Tami's branch

PR #259, opened 2021-10-27
Disabled action when no support

Elysium1436:
When using the unstable branch of neovim, the autocmd CursorHold (on init.lua) was emitting an error when a server that didn't had the capability of code action. 

I added a function that checks if there was a client with the codeAction capability, and used it on other functions that emitted codeAction.


Elysium1436:
I cloned it a while back, wished I knew before wasting time reading the code haha

PR #273, opened 2022-01-03
Fixed deprecated function

Mattio-cmd:
This change should make `:Lspsaga diagnostic()` work properly

### DIFF
--- a/lua/lspsaga/diagnostic.lua
+++ b/lua/lspsaga/diagnostic.lua
@@ -32,7 +32,7 @@ function M.lsp_jump_diagnostic_next(opts)
   return _iter_diagnostic_move_pos(
     "DiagnosticNext",
     opts,
-    vim.lsp.diagnostic.get_next_pos(opts)
+    vim.diagnostic.get(opts)
   )
 end
 


lua/lspsaga/diagnostic.lua
```diff
@@ -32,7 +32,7 @@ function M.lsp_jump_diagnostic_next(opts)
   return _iter_diagnostic_move_pos(
     "DiagnosticNext",
     opts,
-    vim.lsp.diagnostic.get_next_pos(opts)
```
Mattio-cmd, line 35: This is deprecated and should not be used

lua/lspsaga/diagnostic.lua
```diff
@@ -32,7 +32,7 @@ function M.lsp_jump_diagnostic_next(opts)
...
+    vim.diagnostic.get(opts)
```
Mattio-cmd, line 35: As nvim suggest, if you run `:h deprecated` you can see this is the correct way

tami5:
fixed in https://github.com/tami5/lspsaga.nvim

Mattio-cmd:
> fixed in https://github.com/tami5/lspsaga.nvim

Yes, im now using your fork instead of this one. Thanks :)

PR #290, opened 2022-06-19
fix: using lib close_preview_autocmd everywhere

EthanJWright:
vim.lsp.util is throwing an error for me as it does not have close_preview_autocmd. Perhaps this was deprecated?

This PR just uses the utility you created everywhere, so it doesn't break in some places based on the vim api.

An example of the bug I was getting before this PR:

```
E5108: Error executing lua ...ethan/.vim/plugged/lspsaga.nvim/lua/lspsaga/provider.lua:509: attempt to call field 'close_preview_autocmd' (a nil value)                                                                                                                                                                                                                                     
stack traceback:                                                                                                                                                                                                                                                                                                                                                                            
        ...ethan/.vim/plugged/lspsaga.nvim/lua/lspsaga/provider.lua:509: in function 'preview_definition'                                                                                                                                                                                                                                                                                   
        [string ":lua"]:1: in main chunk 
```

### DIFF
--- a/lua/lspsaga/implement.lua
+++ b/lua/lspsaga/implement.lua
@@ -57,7 +57,7 @@ function implement.lspsaga_implementation(timeout_ms)
     }
 
     local bf,wi = window.create_win_with_border(content_opts,opts)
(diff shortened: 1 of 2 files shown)


glepnir:
@EthanJWright  Please try the features branch.

Issue #293, opened 2022-06-23
Open lsp_finder in tab


leoatchina:
**Is your feature request related to a problem? Please describe.**
Open lsp_finder results in tab, like using `tabe`

**Describe the solution you'd like**
Acctually I have folked another `lspsaga.nvim` from `tami5`,  and have added this feature
https://github.com/tami5/lspsaga.nvim/commit/978b2f5845fed65953833b3c3ffd268140b90e0a

glepnir:
if you had implemented.pr welcome

PR #294, opened 2022-06-23
add open finder results in tab, and add a resp == nil judge

leoatchina:
#293 feature added


glepnir:
any update about this pr?

PR #295, opened 2022-06-23
fix: rename

ryanbyon:
- fixed rename (undefined highlight group "CursorWord")
- updated rename section of readme.md for changed function name

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,7 +132,7 @@ src="https://user-images.githubusercontent.com/41671631/105969051-c7fb7700-60c2-
 
 ```lua
 -- rename
-nnoremap <silent>gr <cmd>lua require('lspsaga.rename').rename()<CR>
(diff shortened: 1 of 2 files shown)


ryanbyon:
Hi, thanks for the amazing plugin.

I found a bug that produce error message when trying to rename only when the highlight group "CursorWord" is not set.
In `rename.lua:76.  if vim.fn.hlexists('CursorWord') then`, the function call `hlexists("CursorWord")` return `0` if the user does not use any cursorword highlighting plugins, but lua treats integer `0` as a `true` value, which makes it false positive conditional statement. So the simple comparison solved the issue.

This is my first time contributing to any kind of open source projects, so please understand if I did not follow proper way to contribute.

Thank you :))

glepnir:
No worry it's ok. but when you next create PR you should be use `git checkout -b your-branch`  not use main branch :)

PR #296, opened 2022-06-23
Improve rename

glepnir:
use normal buffer instead of prompt-buffer. in normal buffer we can into visual mode ,I think this is useful . and there has a new option `rename_into_visual` default is true. if set false. won't into visual mode in rename window. remove `rename_prmpt_prefix` option.


glepnir:
rename

PR #298, opened 2022-06-23
fix content with newlines

PHSix:
When I use work for tsx/jsx, if I meet a error and I press <leader>j(which is map to command `:Lspsaga diagnostic_jump_prev`). It will throw a error message(connot contains newlines).

With debug print,I found that `typescript-language-server` will response to a message which the line maybe contains `\n` special character and `nvim_buf_set_lines` api not allow a string contains `\n`character.

before this commit:
without
after this commit:
with


glepnir:
Thx for pr

PR #304, opened 2022-06-26
Add augroup for autocmd created by lspsaga

milanglacier:
autocmd created by `nvim_create_autocmd` using a lua function doesn't have a good string representation:
if you want to know what autocmds you have in an event, like `:autocmd CursorHoldI`,
you will only see a `lua function`, this makes debugging becomes much more difficult if you have some unexpected behavior which might comes from autocmd, if autocmd comes from an anonymous group, this thing becomes worse: you completely have no idea where this autocmd comes from and don't know how to debug.
I add augroup for all the autocmd created by `lspsaga` to mitigate this kind of issues as it is at least better than an anonymous group.

Screen Shot 2022-06-26 at 03 42 48


glepnir:
LGTM!

PR #313, opened 2022-06-27
add `alter` param for smart_scroll_with_saga

leoatchina:
In my [LEOVIM](https://github.com/leoatchina/leovim/blob/master/runtime/lua/lsp-search.lua#L124), I would like to using `<C-j>/<C-k>` as `%/g%` ， so I add a `alter` param instead of  `<C-f>/<C-b>` in the origin function body


glepnir:
what about use `if alter then` not remove line

leoatchina:
OK,  I will update the param

PR #333, opened 2022-07-05
fix(symbolwinbar): index a nil value

JMarkin:
https://github.com/glepnir/lspsaga.nvim/issues/332


nullchilly:
`#node.children > 0`remember that getting size of a table is O(log)

I think you should use `next(node.children)` instead

JMarkin:
> `#node.children > 0`remember that getting size of a table is O(log)
> 
> I think you should use `next(node.children)` instead

Thanks, I don't know about it... I changed it.

PR #335, opened 2022-07-06
refactor(symbolwinbar): Minor performance improvement

nullchilly:
I think `(left + right) >> 1` is better here

`#a > 0` is likely optimized by luajit but optimizing it doesn't make the code uglier so might as well...


glepnir:
LGTM

glepnir:
yeah maybe good .Oh  there has some change in symbolbar 

1 . currently it will ignore user `winbar` config .so provide a function then can set into winbar config. 

2. better events logic .

PR #343, opened 2022-07-07
add option to right-align winbar symbols

WilsonOh:
Hi, sorry if this is not the appropriate branch for this pull request, but I thought it would be nice to have the option to right-align the winbar symbols.
image

### DIFF
--- a/lua/lspsaga/init.lua
+++ b/lua/lspsaga/init.lua
@@ -40,6 +40,7 @@ saga.config_values = {
   symbol_in_winbar = {
     in_custom = false,
     enable = false,
+    right_align = false,
     separator = ' ',
     show_file = true,
     click_support = false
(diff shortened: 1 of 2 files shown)


nullchilly:
Probably a nice option to have
However, I think in order to take full advantage it would be better to use [lspsaga#in_custom](https://github.com/glepnir/lspsaga.nvim#symbolbar-with-your-custom-winbar)
Since you can put the filename on the right and symbols on the left if you want or hide the filename when there is only one window

glepnir:
@WilsonOh  Thanks for pr. and nice job. but we have `in_custom` now. so it can work in any `winbar` plugin or custom winbar config. I think keep easy will be good. so sry I can't merge this pr.

glepnir:
thanks for understand and for your job in lspsaga.

PR #344, opened 2022-07-07
fix(symbolwinbar): Fix highlight

rewhile:
The current way of highlighting is broken

- The `nvim-web-devicons` function is incorrect and won't work for languages. Ex. haskell

- Highlight colors are shared between splits. If we are using devicons anyway you should let users config from there.

### DIFF
--- a/lua/lspsaga/definition.lua
+++ b/lua/lspsaga/definition.lua
@@ -61,12 +61,8 @@ function def.preview_definition(timeout_ms)
       start_line = range.start.line
     end
 
-    local content = vim.api.nvim_buf_get_lines(
-      bufnr,
-      start_line,
-      range['end'].line + 1 + config.max_preview_lines,
-      false
-    )
+    local content =
+      vim.api.nvim_buf_get_lines(bufnr, start_line, range['end'].line + 1 + config.max_preview_lines, false)
     content = vim.list_extend({ config.definition_preview_icon .. 'Definition Preview: ' .. short_name, '' }, content)
 
     local opts = {
--- a/lua/lspsaga/implement.lua
+++ b/lua/lspsaga/implement.lua
@@ -36,12 +36,8 @@ function implement.lspsaga_implementation(timeout_ms)
       start_line = range.start.line
     end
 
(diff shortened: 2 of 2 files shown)


nullchilly:
Maybe expose this function so users can use it in in_custom?

We can also provide an option to enable file_name with dir. Ex `lua > lspsaga > codeaction.lua`

I don't think many will be willing to code these by themselves

glepnir:
I provide `winbar_file_format` before. but I removed. I think this should support by some statusline plugin or themself. 

> I don't think many will be willing to code these by themselves

yup lots of vimers just use neovim. they don't spend some time in config neovim. so make it easy will get more users.

But my philosophy is to keep it simple at the core. Open interfaces provide more support. example of my galaxyline lost a lot of users even though it was the smallest and fastest first statusline plugin to use lua

`winbar` is like `statusline` it will work anywhere in future. Even though we do a lot of things on `winbar`, when 0.8 is released you will see all kinds of `winbar` plugins. we just provide a api that can make symbols show in them winbar.

btw I will support winbar in galaxyline later :)

nullchilly:
Someone will blow the winbar competition in the future yeah...

For the time being I think it wouldn't hurt to provide update readme on `in_custom` on how to accomplish `.config > nvim > plugins.lua`

glepnir:
yup and we can do this  in README at `symbolbar with your custom winbar`

nullchilly:
Thanks, I will open a pr

nullchilly:
@glepnir I forgot to add check f_hl

https://github.com/glepnir/lspsaga.nvim/blob/main/lua/lspsaga/symbolwinbar.lua#L22

Now Telescope is spit errors

My bad but we really need a plenary test so these silly things won't happen :<

glepnir:
simple example .
```lua
function test()
 local is_windows = vim.loop.os_uname().sysname == "Windows" 
 local sep = is_windows and '\\' or '/' 
local path_list = vim.split(vim.fn.expand('%:h'),sep)
 .. then generate winbar string .
end
```

glepnir:
I had add it .

PR #346, opened 2022-07-07
chore(readme): show file with directory

nullchilly:


glepnir:
yup. if still too long .may be add it into wiki page will be better :)

nullchilly:
@glepnir I don't understand why I have to `:redrawtabline` in order for it to work?

https://user-images.githubusercontent.com/56817415/177799306-dbdcb02f-e90d-4d4c-9270-bc3a6cbcb7d9.mp4

nullchilly:
Pretty hacky but it works?

I added `vim.cmd "doautocmd CursorMoved"` inside lsp callback in the above commit

glepnir:
looks like this code works fine with `in_custom`

```lua
  api.nvim_create_autocmd({ 'TextChanged', 'InsertLeave' }, {
    group = saga_group,
    buffer = current_buf,
    callback = function()
      if config.in_custom then
        symbar:get_buf_symbol(true)
      else
        symbar:get_buf_symbol(true, render_symbol_winbar)
      end
    end,
    desc = 'Lspsaga update symbols',
  })
```
template

nullchilly:
If you do nothing after lspsaga received LSP request there is no way to call user's in_custom command

glepnir:
this is expect.because we register events in lspattach . this event is triggered when lsp server attach that mean it will happen after bufwinenter
so when you move it will show. add the hack that mean manually trigger curosr move to render . cursorhold etc also work

nullchilly:
I mean adding the hack makes `get_buf_symbol` callback in_custom user-defined winbar so why not?

Or I can just make LspsagaAttach autocmd

glepnir:
oh i just do update cache in these command，when user edit the code ，update cache，then if cursor move it，it will read new cache and show it in，winbar，if not move winbar still show old cache symbol，problem is this right？

nullchilly:
Yes

glepnir:
try add cursorhold in custom winbar events.

nullchilly:
It doesn't work, it posted the reasons up there

glepnir:
Okay test it later，maybe add a param in get symbol function if true.do cursor moved.sry i use github on my phone

glepnir:
wait just use in_custom

nullchilly:
```
api.nvim_create_augroup('LspsagaAttach', {})
```

I tried to add a custom autocmd but it doesn't work...

```
  local _callback = function(results)
    if libs.result_isempty(results) then
      return
    end

    self.symbol_cache[current_buf] = { true, results[client_id].result }

		vim.cmd "doautocmd LspsagaAttach"

    if fn ~= nil then
      fn()
    end
  end

  do_symbol_request(_callback)
```

glepnir:
I know how to do it，will test later

nullchilly:
I think just CursorHold is good enough.

glepnir:
don't do that I think. it will trigger all function of `CursorMoved`

nullchilly:
I put CursorHold not CursorMoved

Either way create LspsagaAttach is still better I guess

nullchilly:
@glepnir I have reverted the doautocmd for now since I don't have much free time left

Can you merge the click support fix?

glepnir:
use `User LspasgaUpdateSymbol` in autocmds.

nullchilly:
There seems to be some upstream changes about `g.did_load_filetypes = 0 g.do_filetype_lua = 1` and I had no filetype detection for the last 15 minutes...

glepnir:
yes it's in default , no need set it in your config

nullchilly:
@glepnir Please review

glepnir:
I think we don't need these

```lua
    if config.in_custom then
      vim.cmd('doautocmd User LspasgaUpdateSymbol')
    end
```

nullchilly:
Hmm lemme check

nullchilly:
Yeah we don't

glepnir:
there has a element channel in readme you can join it  for quick reply

PR #351, opened 2022-07-08
fix(symbolwinbar): Extra space on non-icon buffer

nullchilly:
Also, the current winbar behavior is not very good
If `vim.wo.winbar = nil` then it will not remove the extra one line. However `vim.wo.winbar = ""` does... Upstream bug?

https://user-images.githubusercontent.com/56817415/177960834-8f1efa04-cb42-4dc7-b179-5fff209687ec.mp4

in_custom manage to overcome this because I pattern matching every buffer

https://user-images.githubusercontent.com/56817415/177961447-5c8a2106-57df-4712-b3b6-4facdac0c218.mp4

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,7 +151,8 @@ to get symbols node and set `User LspsagaUpdateSymbol` event in your autocmds
 -- Example:
 local function get_file_name(include_path)
     local file_name = require("lspsaga.symbolwinbar").get_file_name()
-    if include_path == false then return require("lspsaga.symbolwinbar").get_file_name() end
+    if vim.fn.bufname("%") == "" then return "" end
+    if include_path == false then return file_name end
     -- Else if include path: ./lsp/saga.lua -> lsp > saga.lua
     local path_list = vim.split(vim.fn.expand('%:~:.:h'), vim.loop.os_uname().sysname == "Windows" and '\\' or '/')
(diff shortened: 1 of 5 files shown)


glepnir:
`:h winbar` you can't set it to nil. default is empty

glepnir:
hmm works fine with me ?

image

nullchilly:
@glepnir https://github.com/neovim/neovim/issues/18660 upstream issue

glepnir:
I think it should be work like statusline/tabline.

nullchilly:
Sure, I think it's a preference

Either way, I provided a way to disable it if they want in this pull request. Please review

nullchilly:
Thanks @glepnir

I saw you joined the neovim core, wish you luck for the foreseeable :)

Taking a break from the internet for a while, thanks for merging these silly pull requests!

glepnir:
oh I had joined for a while . now i public my organizations team information. so you saw it :)

 There are some code not very readable in example . I had edit it . a reusable table should not be defined inside a function. Memory allocation is unscientific

glepnir:
@nullchilly  hi did you found symbol winbar very slow in lua file today?

nullchilly:
@glepnir It's faster actually lol

PR #353, opened 2022-07-10
fix split_by_pathsep

kaizhang16:
`split_by_pathsep` can't be found after https://github.com/glepnir/lspsaga.nvim/commit/9922e17c8e535328462fa246cc8fde32d9f72371, so update definition.lua as well.


glepnir:
Oh I forget it used in definition. but ci failed. you should fix format by stylua

PR #359, opened 2022-07-15
feat: improve finder ui

### LINKED ISSUE #355: Lsp_finder showing incorrect number of definitions and rendering incorrectly
### Description

I'm using lspsaga on csharp file, with omnisharp-roslyn as the lsp server. There is only one definition for `PaginatedResponse`, but lsp_finder shows that it has found 2 definitions.

Briefly looking into the issue, it seems that within finder.lua when creating the `results`, lspsaga is inadvertenly splitting up the range and uri objects to create two entries under the definitions.

The response coming back from omnisharp-roslyn looks ok, but when looking at the table for the response results using `vim.inspect` `textDocument/definition` looks like this:

`result = { range = {} uri = "" }`

It seems like the code for generating the results is expecting to have multiple values within result, as the response results for `textDocument/references` instead look like:

`result = { { range = {}, uri = {} }, { range = {}, uri = {} }`

**Expected Behavior**

lsp_finder shows the correct number of definitions, renders correctly. Something like this:

image

**Actual Behavior**

lsp_finder is showing 2 definitions (there is only 1) and it seems to be causing the window to render things incorrectly:

image

I've created a work-around locally by checking that `table.getn(res.result) == 0` and if so, just inserting `res.result` straight into the results. The result of this workaround you can see above in the expected behaviour. Obviously, I have no idea if this is appropriate place for a fix so just wanted to raise an issue to see if this is something others have seen or if it's a quirk with my particular setup.

<details><summary>Environment</summary>

- nvim --version output: 
NVIM v0.7.0
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compiled by runneradmin@fv-az320-113
- Operating system: 
Windows 10
- lspsaga commit: 
Latest main

</details>

glepnir:
improve finder ui and fix #355 

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,8 +98,12 @@ code_action_lightbulb = {
     sign_priority = 20,
     virtual_text = true,
 },
--- separator in finder
-finder_separator = "  ",
+-- finder icons
+finder_icons = {
+  def = '  ',
+  ref = '諭 ',
+  link = '  ',
+},
 -- preview lines of lsp_finder and definition preview
 max_preview_lines = 10,
 finder_action_keys = {
@@ -254,6 +258,7 @@ The available highlight groups you can find in [here](./plugin/lspsaga.lua).
 <details>
 <summary>Async lsp finder</summary>
 
+Finder Title work with neovim 0.8 +
 
 **Lua**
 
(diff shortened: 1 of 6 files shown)


glepnir:
finder

PR #388, opened 2022-07-31
fix(lightbulb): use passed in bufnr

Littledarren:
the current buffer may have been changed during coroutine running and that causes an error.

### DIFF
--- a/lua/lspsaga/lightbulb.lua
+++ b/lua/lspsaga/lightbulb.lua
@@ -1,7 +1,7 @@
 local api = vim.api
 local libs = require('lspsaga.libs')
 local config = require('lspsaga').config_values
-local method = 'textDocument/codeAction'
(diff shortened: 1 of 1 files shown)


glepnir:
LGTM

XXiaoA:
> the current buffer may have been changed during coroutine running and that causes an error.

Hi, after merging your PR, nvim threw the error everytime when I'm in `i` mode and move the cursor
img

glepnir:
`sign_unplace` can't use  0, `0` it's a highly function in neovim api-fast.

Littledarren:
Sorry for the error. I passed bufnr= 0 and want it to be the default current buffer num which fails. Need I give a quick patch or pr?

Littledarren:
I've seen the fix, sorry for the bug...抱歉hh

Issue #432, opened 2022-08-14
Using add-import code action in haskell performs wrong action


mizlan:
### Description
I am in a Haskell file, and if there is a symbol I want to import, and I try using LspSaga code action, it always picks this different option that is not very helpful (will explain in more detail below).

**Expected Behavior**
The "add import" action is executed (e.g., `import Data.Maybe` is added to top of file)

**Actual Behavior**
For Haskell, I have `{-# OPTIONS_GHC -Wno-deferred-out-of-scope-variables #-}` added to the top of my file (which is another codeaction option)

**Neovim Built in Behavior**
`vim.lsp.buf.code_action` works fine

### Details

<details><summary>Reproduce</summary>

1. nvim -u test.vim the min config that only lspsaga
```lua
P = require('neopm')

P 'https://github.com/neovim/nvim-lspconfig'
P 'https://github.com/glepnir/lspsaga.nvim'
P.autoinstall(true)
P.load()

local on_attach = function(client, bufnr)
  local bufopts = { noremap = true, silent = true, buffer = bufnr }
  vim.keymap.set('n', '<leader>ca', "<cmd>Lspsaga code_action<CR>", bufopts)
end
require('lspconfig')['hls'].setup {
  on_attach = on_attach,
}
```
2. post your program language min code snippet that can reproduce your issue

1 line:
```haskell
c = catMaybes
```
There is error on catMaybes, variable not in scope, but if you use codeAction add import it doesnt work, instead the long GHC option is added.

3. instead of that, `import Data.Maybe` should be added.
</details>

<details><summary>Environment</summary>

- nvim --version output: 
```
NVIM v0.8.0-dev+608-g778541067
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by ml@Michaels-MacBook-Pro.local

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/opt/homebrew/Cellar/neovim/HEAD-7785410/share/nvim"
```
- Operating system: macOS monterey
- lspsaga commit: f4041254c242dee33cd14a047471b2abc1fe2fa2

</details>

glepnir:
`{-# OPTIONS_GHC -Wno-deferred-out-of-scope-variables #-}`  put this line at the head of file?

mizlan:
Lspsaga puts that line there even though thats not the option I want.

glepnir:
[...]

PR #441, opened 2022-08-15
fix: allow user to pass data into codeaction API

mizlan:
Previously, all data that was passed as arguments into one of these
functions would be discarded, since these functions are defined to take
no arguments.

This makes it so that these outer API functions pass all arguments
directly to the inner functions. This allows the user to for example
press "2" and actually have "2" transmitted to the inner function,
instead of being lost.

This also may likely prevent a whole host of other potential bugs.


mizlan:
Fixes #432

PR #456, opened 2022-08-23
remove signaturehelp

glepnir:


quantumsnowball:
Hi there, I found this merged PR until I update all repos in Packer and a nil exception appeared. I have been using signaturehelp in Lspsaga. May I ask why removing this feature? And are there planning to add back a similar feature? Thanks.

glepnir:
@quantumsnowball  sorry for that . you can use `vim.lsp.buf.signaturehelp` instead of currently. why I remove it. because I think it should be works with completion . when completion a function it insert function name with brackets then trigger it .
this is what I don in my custom completion plugin :) not stable now

PR #462, opened 2022-08-27
fix(kind): `custom_kind` wouldn't take effect in some occasions.

Jint-lzxy:
## Current behavior
`custom_kind` would only be respected when symbols in winbar is enabled:
https://github.com/glepnir/lspsaga.nvim/blob/b59ff0d1752f7e9ef93363b71a2cd9b8f2011fec/lua/lspsaga/init.lua#L105-L108
Where `require('lspsaga.lspkind').gen_symbol_winbar_hi()` modified lspkind icons.

## Expected behavior
`custom_kind` would overwrite icons / colors in any occasion.

### Solution
Handle the `custom_kind` table separately.

### DIFF
--- a/lua/lspsaga/lspkind.lua
+++ b/lua/lspsaga/lspkind.lua
@@ -56,27 +56,27 @@ local function find_index_by_type(k)
   return nil
 end
 
+if next(custom_kind) ~= nil then
+  for k, conf in pairs(custom_kind) do
+    local index = find_index_by_type(k)
+    if not index then
+      vim.notify('Does not find this type in kind')
+    end
+
+    if type(conf) == 'string' then
+      kind[index][3] = conf
+    end
+
+    if type(conf) == 'table' then
+      kind[index][2] = conf[1]
+      kind[index][3] = conf[2]
+    end
+  end
+end
+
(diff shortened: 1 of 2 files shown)


lua/lspsaga/lspkind.lua
```diff
@@ -56,27 +56,27 @@ local function find_index_by_type(k)
...
-  local prefix = 'LspSagaWinbar'
-  local winbar_sep = 'LspSagaWinbarSep'
-
-  if next(custom_kind) ~= nil then
-    for k, conf in pairs(custom_kind) do
-      local index = find_index_by_type(k)
-      if not index then
-        vim.notify('Does not find this type in kind')
-      end
+if next(custom_kind) ~= nil then
```
glepnir, line 59: you made if outside . so this statement will be effect when lspkind.lua require.  should be set into a init kind funciton. then require it in saga init.

```lua
function load_custom_kind()
  if next(custom_kind) == nil then
     return
  end
 do some stuff
end

```

then in init_saga function require this function .

Jint-lzxy: Fixed :)

PR #477, opened 2022-08-30
fix(rename): close the window if is normal mode after pressing quit_key

XXiaoA:
And, if `rename_action_quit` is `<C-c>` it will close in all mode

### DIFF
--- a/lua/lspsaga/rename.lua
+++ b/lua/lspsaga/rename.lua
@@ -35,7 +35,9 @@ function rename:apply_action_keys()
 
   for i, mode in pairs(modes) do
     vim.keymap.set(mode, quit_key, function()
-      self:close_rename_win()
+      if vim.fn.mode() == 'n' or quit_key == '<C-c>' then
+        self:close_rename_win()
+      end
     end, { buffer = self.bufnr })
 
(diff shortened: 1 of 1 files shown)


lua/lspsaga/rename.lua
```diff
@@ -35,7 +35,9 @@ function rename:apply_action_keys()
 
   for i, mode in pairs(modes) do
     vim.keymap.set(mode, quit_key, function()
-      self:close_rename_win()
+      if vim.fn.mode() == 'n' or quit_key == '<C-c>' then
```
glepnir, line 38: what does this mean? why check mode in function .it map to `normal insert visual` mode by default.

XXiaoA: I comfigure quite_key to`q`, But this map doesn't check the mode. So I can't type any word has character "q". Also, we can press `<C-c>` to quit in all mode

glepnir: doesn't make，sense I don't bind it to q  is useful .if function.name or variable has q in new name .what does this do? and check map in there? if other people use ctrl d . what does this patch do in check ctrl c?

PR #487, opened 2022-08-31
DONT MERGE THIS STUPID PR INTO MAIN!!!!!

NEX-S:
This is a stupid PR.
Though I'm the freshman for neovim, so the code in this PR is terrible, but I wish can improve lspsaga as best as I can.

I ADD some stuff might be useful, but the code is definitely not the best, some implements is stupid.

New feature:
1. ADD float term config
```lua
    float_term = {
        winblend = 25,                                 -- winblen of float_term
        shadow_background = false,           -- wether show dark background
    },
```

2. definition preview quit key config
```
definition_preview_quit = '<C-c>',      -- default to <C-c>
```

3. autojump on symbol in preview window
this is implements by this
```
local res = vim.fn.expand('<cword>')
open definition preview window
vim.fn.search(res) to jump on result
...
```

4. auto highlight finder's result without affect any native register
this is implements by
```
let g:search = @/
set hls
let @/ = expand('<cword>')
open_finder with result highlight in search
...
close_finder with let @/ = g:search
set nohls
```

BUGS:
1. if float_term is exit by user not hind by lspsaga, re open it will cause error (so we might need two option like open_new / resume floating window)

It's just my idea, I'm just trying to improve lspsaga's experience ( the way I implement it's not the best way but I keep trying)

Hope it can help you, even if it reduces the workload a little, wish you have a good day sir :100: 

best wishes

### DIFF
--- a/lua/lspsaga/definition.lua
+++ b/lua/lspsaga/definition.lua
@@ -73,7 +73,7 @@ function def:preview_definition()
     )
 
     local prompt = config.definition_preview_icon .. 'File: '
-    content = vim.list_extend({ prompt .. short_name .. ' [Ctrl + c] to quit', '' }, content)
+    content = vim.list_extend({ prompt .. short_name, '' }, content)
 
     local opts = {
       relative = 'cursor',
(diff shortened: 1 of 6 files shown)


glepnir:
no need use this way

PR #491, opened 2022-09-01
Allow quitting preview_definition with custom shortcut

vihu:
Fix https://github.com/glepnir/lspsaga.nvim/issues/490

### DIFF
--- a/lua/lspsaga/definition.lua
+++ b/lua/lspsaga/definition.lua
@@ -71,7 +71,7 @@ function def:preview_definition()
 
     local prompt = config.definition_preview_icon .. 'File: '
     local quit = config.definition_action_keys.quit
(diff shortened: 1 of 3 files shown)


lua/lspsaga/definition.lua
```diff
@@ -70,7 +70,7 @@ function def:preview_definition()
     )
 
     local prompt = config.definition_preview_icon .. 'File: '
-    content = vim.list_extend({ prompt .. short_name .. ' [Ctrl + c] to quit', '' }, content)
+    content = vim.list_extend({ prompt .. short_name .. ' q to quit', '' }, content)
```
glepnir, line 73: if you make it configurable .there also need use that variable.

PR #508, opened 2022-09-10
ADD `defnition_action_keys.enter` to quick jump into definition when peek_definition

NEX-S:


glepnir:
0.2.2 version `peek_definition` is support you jump into it and it's a normal buffer .so you can do edit save etc.

glepnir:
it had implement some actions .will update later.

PR #520, opened 2022-09-14
fix(definition): support peeking multiple definitions

Alienover:
1. Peeking definitions within the same buffer
- [x] Only remove the keymappings when all the related preview windows are closed
- [x] Keep the definition windows relationships based on the buffer ID

2. Peeking definitions across buffers/files
- [x] Only remove the keymappings for the closing buffer

3. edit/split/vsplit/tabe
- [x] Close all the windows and non-related buffers when triggering edit/split/vsplit/tabe

https://user-images.githubusercontent.com/12182069/190087724-c065774a-b551-4e40-8a79-c1e7315e716f.mov


glepnir:
can you add usage in README -> example config section .

Alienover:
> can you add usage in README -> example config section .

This PR works without any config changed.

PR #524, opened 2022-09-16
finder: allow tables for all action keys

tobealive:
Hello @glepnir, 

What a great plugin!
I had it on the list to implement it since you had your "come back" and started coding again 😊 I hope you are doing fine! I was using a lot of custom stuff in the meantime. Finally, I could make time to implement it this week. It's outstanding work done with this plugin! Thanks for sharing it 🙏

Maybe you can make use of this PR. I think as it extends functionality while simultaneously simplifying code it could be interesting.

### DIFF
--- a/lua/lspsaga/definition.lua
+++ b/lua/lspsaga/definition.lua
@@ -172,25 +172,35 @@ function def:apply_aciton_keys(scope, bufnr, pos)
   local maps = config.definition_action_keys
 
   for action, key in pairs(maps) do
-    vim.keymap.set('n', key, function()
-      local curr_scope = self:find_current_scope()
-      local link, def_win_ns = curr_scope.link, curr_scope.def_win_ns
+    local function map_action(keymap)
+      vim.keymap.set('n', keymap, function()
+        local curr_scope = self:find_current_scope()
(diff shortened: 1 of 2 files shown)


tobealive:
The idea for extending keymaps to accept tables originally came from having problems with moving out of the `definition-window`. 

Out of muscle memory I would often move out of it using my "window-switch" keymaps. Which resulted in the definition-window staying open, and I could not move back in to close it. To also solve this I additionally wanted to map my "window-switch" keymaps to the quit action, which would not work currently as only a string is accepted.

To also allow this, I'll add the commit to extend `defintion_action_keys` to have tables. As a slightly different approach is used here. I hope I got it all right and to your liking.

glepnir:
If you want each action key support string or table.  why not abstract. you write a lot of ifelse which looks bad.

```lua
local key_type = {
  ["table"] = function()
  end,
 ["string"] = function()
  end,
}

local _mt = { }

_mt:index = function(_,type)
  vim.notify("type are not support")
end

key_type[type(action_key)]()
```

glepnir:
maybe I will implement it in 0.2.3 there has a new ui for nvim 0.8

PR #538, opened 2022-09-25
fix: fix rename bug

LimbicSys:
If the treesitter parser is not installed, return true directly.

### DIFF
--- a/lua/lspsaga/rename.lua
+++ b/lua/lspsaga/rename.lua
@@ -106,7 +106,13 @@ local function support_change()
   local bufnr = api.nvim_get_current_buf()
   local queries = require('nvim-treesitter.query')
   local ft_to_lang = require('nvim-treesitter.parsers').ft_to_lang
-  local query = queries.get_query(ft_to_lang(vim.bo[bufnr].filetype), 'highlights')
+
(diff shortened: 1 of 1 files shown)


lua/lspsaga/rename.lua
```diff
@@ -106,8 +106,14 @@ local function support_change()
   local bufnr = api.nvim_get_current_buf()
   local queries = require('nvim-treesitter.query')
   local ft_to_lang = require('nvim-treesitter.parsers').ft_to_lang
-  local query = queries.get_query(ft_to_lang(vim.bo[bufnr].filetype), 'highlights')
-
+  
+  local lang = ft_to_lang(vim.bo[bufnr].filetype)
+  local is_installed = #api.nvim_get_runtime_file("parser/" .. lang .. ".so", false) > 0
```
glepnir, line 111:
```suggestion
 local is_installed = #api.nvim_get_runtime_file('parser/' .. lang .. '.so', false) > 0
```

LimbicSys: Reformatted.

glepnir:
LGTM ! Thanks.

PR #540, opened 2022-09-28
Support custom winbar filename

leaxoy:
It's convenient to custom winbar filename with custom file formatter.
When set file_formatter to `%:.`, it will expand to:
image


glepnir:
should be document.

leaxoy:
Done

PR #581, opened 2022-11-04
ADD: display lines above in lsp_finder

hsnks100:
image


hsnks100:
to pass stylua, fix some codes. 

"table" -> 'table' in definition.lua

PR #622, opened 2022-12-17
Fix outline folded text

po1o:
In function set_foldtext(), the foldtext is a concatenation of:
'outline_conf.fold_prefix' and the line at foldstart.
By default, fold_prefix is undefined and an empty string is returned.
This change defines fold_prefix in the default config.


glepnir:
sry I had refactor outline in 0.2.3 see pr

po1o:
Sounds good! Thanks a lot. 

I can live with a local fix to my config in the meantime ;-)

PR #631, opened 2022-12-29
fix: when lsp reference the file empty report error and render a hint “the file is empty”

jinzhongjia:
when use lsp reference to find function， if the file is empty, nvim will report error
图片

I added a piece of judgment logic to render the text the file is empty when the file is empty, and the highlight function will not be executed if the file is empty

图片


jinzhongjia:
I haven't read the entire finder logic, so I don't know if there is redundancy in my diagnostic modification

lua/lspsaga/finder.lua
```diff
@@ -626,7 +626,21 @@ function Finder:auto_open_preview()
   if not self.short_link[current_line] then
     return
   end
-  local content = self.short_link[current_line].preview or {}
+
+  local content
+  local table_length = 0
+  for key, value in pairs(self.short_link[current_line].preview) do
```
glepnir, line 632: there can be `vim.tbl_count(self.short_link[current_line].preview`

jinzhongjia: ok, i will change it

jinzhongjia:
all right, i have changed it

PR #642, opened 2023-01-13
Fix some little grammar mistakes in init_lsp_saga

HendrikPetertje:
The changed message gets rid of some grammar mistakes and makes it a bit clearer what is expected of the user (migrate to a specific new callable "setup" function)

--- 

Thank you so much for the amazing work you are doing on this plugin, I have been enjoying it for some years now already and I really like the stuff this plugin does!

And thanks for making these huge changes to it now and streamlining the config! I don't entirely agree with the styling of the float windows and their colors (I just like it default as managed by my editor's theme with lines being the same color as my indent-lines and text being a bit less flashy) but you included a nice list of theming options already, so I'll start trying those out :)

winbar seems to work a lot better (even when called through my lualines config) and many things have received subtle, but awesome improvements. So thanks a bunch!

### DIFF
--- a/lua/lspsaga/init.lua
+++ b/lua/lspsaga/init.lua
@@ -139,7 +139,7 @@ end
 ---@deprecated
 function saga.init_lsp_saga()
   vim.notify(
-    'lspsaga.nvim version 0.2.3 is breakchange please read doc and use setup function!',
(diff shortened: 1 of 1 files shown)


HendrikPetertje:
jep, experimenting with that now! thanks so much for maintaining and working on this plugin!

PR #688, opened 2023-01-15
fix(hover): use the correct types

uga-rosa:
I'm sorry.
I've taken a few more considerations into account and sorted it out.
Related PR;
#674
#685


glepnir:
Thanks.

PR #705, opened 2023-01-16
fix the bug of losing index in diagnostic

YangEfei:
The index of diagnostic window will be dropped since the extmark overlap it.
The extmark will take up 2  char positions, so I set 2 spaces before the index to solve it.

### DIFF
--- a/lua/lspsaga/diagnostic.lua
+++ b/lua/lspsaga/diagnostic.lua
@@ -461,7 +461,7 @@ function diag:show(entrys, arg, type)
     local code_source =
       api.nvim_buf_get_text(entry.bufnr, entry.lnum, start_col, entry.lnum, end_col, {})
     insert(len, #code_source[1])
-    local line = '['
+    local line = '  ['
       .. index
       .. '] '
       .. code_source[1]


lua/lspsaga/diagnostic.lua
```diff
@@ -461,7 +461,7 @@ function diag:show(entrys, arg, type)
     local code_source =
       api.nvim_buf_get_text(entry.bufnr, entry.lnum, start_col, entry.lnum, end_col, {})
     insert(len, #code_source[1])
-    local line = '['
+    local line = '  ['
```
glepnir, line 464: why there is two not one space

YangEfei:
A more visual illustration of the problem.
Before:
image

After:
image

Maybe there's a better way to solve it. Please feel free to fix it.

lua/lspsaga/diagnostic.lua
YangEfei: You can see the screenshot below. The extmark will take up 2 chars.
The '[' and the index number lose.

glepnir:
why i use `conceal` because i want to hide these numbers ..need rewrite there number should keep..

btw does you version of lspaga is latest highlight have some problem . I had fixed it before
image

glepnir:
image

YangEfei:
> why i use `conceal` because i want to hide these numbers ..need rewrite there number should keep..
> 
> btw does you version of lspaga is latest highlight have some problem . I had fixed it before image

image
image

It because I set the transparency in my iTerm2. There's the same problem in lualine and powerline. As a matter of fact, it maybe not a problem I think. If I change the iTerm2 to the full screen, it works fine.
So you mean that you want to ignore these index numbers as fact? I thought it was a feature before. And I can input the number to choose the corresponding code action which I want to execute. 😂

YangEfei:
> image

That looks so good. Just apply it~

glepnir:
also support press number to execute but only code action win.  I can also add this support in diagnostic jump win then you don't need jump into it . wait a while

YangEfei:
image
The ']' seems still here... 😂

glepnir:
and I also guess it's your font ? or something else ...may be you can try to change the `libs.unciode_numbers`.. in my guess is your font.

YangEfei:
It's still there after I pull the latest commit (018473e41aefb572d90b2ca4885be41df9f63bbc).
And I have tired all nerd font in my macOS, contains Meslo, FiraCode and Hack. They all meet the same problem.
It's FiraCode in the screenshot:
image
And I even tried to increase the font size and width, but nothing changed. The ']' is still there. 😢

glepnir:
you don't need install the nerd fonts. nerd font have a font that only have symbols. some modern terminal support set font for special symbols. like wezterm built in support ,  other terminal you just need download the symbols nerdfont then install . like I use Operator mono in normal. symbols use symbol nerd font.

image

YangEfei:
image
I tried Operator Mono just now, and it's still the same as before... 
(The icon can not show because I don't set the font for non ASCII font, but it's same even if I set it. 😓)
BTW, this font is pretty good. 😆

glepnir:
IDK then .

glepnir:
regex also need rewrite.

PR #706, opened 2023-01-16
Fix the error when `respect_root = true`

YangEfei:
When `respect_root = true`, there will be an error due to wrong type of parameter.


YangEfei:
If I set `respect_root = true`, then there will be an error when I open a file every time.
image

PR #723, opened 2023-01-18
Fix hover_doc

yuki-yano:
Related: #674 #688

Fixed an error in tsserver due to incorrectly positioned code for converting contents.

### DIFF
--- a/lua/lspsaga/hover.lua
+++ b/lua/lspsaga/hover.lua
@@ -116,7 +116,7 @@ function hover:do_request(arg)
       return
     end
 
-    if not result or not result.contents or next(result.contents) == nil then
+    if not result or not result.contents then
       if not arg or arg ~= '++quiet' then
         vim.notify('No information available')
       end


glepnir:
just wonder why it behind the parser content.  when request done . we should check result first.   then check result.content after parse. so it should .

```lua
if not result or not result.content then
return
end

parse result.content
....

yuki-yano:
Sorry, I fixed it to work, but I think the code was wrong and I will check it.
However, the current code sometimes has a string in `result.contents`, so I get an error at `next(result.contents)`.

glepnir:
yes . res.content not always table so remove this check `next(result.content) == nil`

yuki-yano:
Thank you. Should I delete this PR?

glepnir:
you can just remove this `or next(result.contents) == nil `

yuki-yano:
Modified PR to remove only the part that uses next.

glepnir:
Thanks

PR #779, opened 2023-01-27
version 0.2.4

glepnir:
- [x] window behavior improve 
- [x] `show_buf/cursor/line_diagnostic` support `<CR>` jump
- [x] support config jump beacon
- [x] refactor highlight

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,6 +303,8 @@ Default options:
     show_code_action = true,
     show_source = true,
     jump_num_shortcut = true,
+    custom_fix = nil,
+    custom_msg = nil,
     keys = {
       exec_action = "o",
       quit = "q",
@@ -313,6 +315,8 @@ Default options:
 
 - Using `go_action`, you can quickly jump to line where actions need to be taken in the diagnostics floating window.
 - `jump_num_shortcut` - The default is `true`. After jumping, Lspasga will automatically bind code actions to a number. Afterwards, you can press the number to execute the code action. After the floating window is closed, these numbers will no longer be tied to the same code actions.
+- `custom_msg` string  used to  custom the diagnostic jump `Msg` section titile 
+- `custom_fix` string  used to  custom the diagnostic jump `Fix` section titile 
 
 You can also use a filter when using diagnostic jump by using a Lspsaga function. The function takes a table as its argument.
 It is functionally identical to `:h vim.diagnostic.get_next`.
--- a/README.md
+++ b/README.md
@@ -344,6 +344,8 @@ The steps demonstrated in this showcase are:
 
 `show_line_diagnostics`, `show_buf_diagnostics`, `show_cursor_diagnostics`
 
+- you can use `<C-w>w` jump into and use `<CR>` jump to diagnostic position
+
 <details>
 <summary>show_diagnostics showcase</summary>
 <img src="https://user-images.githubusercontent.com/41671631/212220793-a52215fd-5f60-4be6-8132-78247b921f1e.gif" height=80% width=80%/>
@@ -493,6 +495,19 @@ A simple floating terminal.
 <img src="https://user-images.githubusercontent.com/41671631/212027060-56d1cebc-c6a8-412e-bd01-620aac3029ed.gif" height=80% width=80%/>
 </details>
 
+## :Lspsaga beacon
+
+after jump from float window there will show beacon to remind you where the cursor is.
+
+```lua
+  beacon = {
+    enable = true,
+    frequency = 7,
+  },
+```
+
+`frequency` the blink frequency.
+
 ## Customizing Lspsaga's Appearance
 
 ## :Lspsaga UI
@@ -514,29 +529,13 @@ Default UI options
     diagnostic = "🐞",
     incoming = " ",
     outgoing = " ",
-    colors = {
-      -- Normal background color for floating window
-      normal_bg = "#1d1536",
-      -- Title background color
-      title_bg = "#afd700",
-      red = "#e95678",
-      magenta = "#b33076",
-      orange = "#FF8700",
-      yellow = "#f7bb3b",
-      green = "#afd700",
-      cyan = "#36d0e0",
-      blue = "#61afef",
-      purple = "#CBA6F7",
-      white = "#d1d4cf",
-      black = "#1c1c19",
-    },
+    hover = ' ',
     kind = {},
   },
 ```
 
 # Custom Highlighting
 
-If you don't like the colors or if you want to switch your colorscheme over time, you can change the highlight group in your colorscheme or config.
 All highlight groups can be found in [highlight.lua](./lua/lspsaga/highlight.lua).
 
 # Custom Kind
--- a/doc/lspsaga.nvim.txt
+++ b/doc/lspsaga.nvim.txt
@@ -571,8 +571,6 @@ Default UI options
 
 >
       ui = {
-        -- Currently, only the round theme exists
-        theme = "round",
         -- This option only works in Neovim 0.9
         title = true,
         -- Border type can be single, double, rounded, solid, shadow.
--- a/lua/lspsaga/callhierarchy.lua
+++ b/lua/lspsaga/callhierarchy.lua
@@ -492,7 +492,7 @@ function ch:preview()
   local rbottom = window.combine_char()['rightbottom'][config.ui.border]
   local content_opt = {
     contents = {},
-    setbuf = data[1],
+    bufnr = data[1],
     border_side = {
       ['lefttop'] = rtop,
       ['leftbottom'] = rbottom,
--- a/lua/lspsaga/callhierarchy.lua
+++ b/lua/lspsaga/callhierarchy.lua
@@ -118,7 +118,7 @@ function ch:call_hierarchy(item, parent)
     if not res or next(res) == nil then
       return
     end
-    local kind = require('lspsaga.highlight').get_kind()
+    local kind = require('lspsaga.lspkind').get_kind()
     if not parent then
       local icons = {}
       for i, v in pairs(res) do
--- a/lua/lspsaga/callhierarchy.lua
+++ b/lua/lspsaga/callhierarchy.lua
@@ -349,27 +349,34 @@ function ch:render_win()
     },
   }
 
+  local cur_winline = fn.winline()
+  local max_height = math.floor(vim.o.lines * 0.4)
+  if vim.o.lines - cur_winline - 6 < max_height then
+    vim.cmd('normal! zz')
+    local keycode = api.nvim_replace_termcodes('5<C-e>', true, false, true)
+    api.nvim_feedkeys(keycode, 'x', false)
+  end
+
   local opt = {
     relative = 'editor',
-    row = math.floor(vim.o.lines * 0.2),
-    col = math.floor(vim.o.columns * 0.1),
+    win = api.nvim_get_current_win(),
+    row = fn.winline() + 1,
+    col = 10,
     height = math.floor(vim.o.lines * 0.4),
     width = math.floor(vim.o.columns * 0.3),
     no_size_override = true,
   }
 
-  if fn.has('nvim-0.9') == 1 and config.ui.title then
-    local theme = require('lspsaga').theme()
+  if fn.has('nvim-0.9') == 1 then
     local icon = self.method == 'callHierarchy/incomingCalls' and ui.incoming or ui.outgoing
     opt.title = {
-      { theme.left, 'TitleSymbol' },
-      { icon, 'TitleIcon' },
-      { ' ' .. self.method:match('/(%w+)Calls$'), 'TitleString' },
-      { theme.right, 'TitleSymbol' },
+      { icon, 'ArrowIcon' },
     }
+    opt.title_pos = 'right'
+    api.nvim_set_hl(0, 'ArrowIcon', { link = 'CallHierarchyBorder' })
   end
   self.bufnr, self.winid = window.create_win_with_border(content_opt, opt)
-  api.nvim_win_set_cursor(self.winid, { 2, 8 })
+  api.nvim_win_set_cursor(self.winid, { 2, 9 })
   api.nvim_create_autocmd('CursorMoved', {
     buffer = self.bufnr,
     callback = function()
@@ -471,15 +478,14 @@ function ch:preview()
     return
   end
 
+  local winconfig = api.nvim_win_get_config(self.winid)
   local opt = {
-    relative = 'editor',
+    relative = winconfig.relative,
+    row = winconfig.row[false],
+    height = winconfig.height,
+    col = winconfig.col[false] + 2 + winconfig.width,
     no_size_override = true,
   }
-
-  local winconfig = api.nvim_win_get_config(self.winid)
-  opt.height = winconfig.height
-  opt.row = winconfig.row[false]
-  opt.col = winconfig.col[false] + winconfig.width + 2
   opt.width = vim.o.columns - opt.col - 6
 
   local rtop = window.combine_char()['righttop'][config.ui.border]
--- a/lua/lspsaga/callhierarchy.lua
+++ b/lua/lspsaga/callhierarchy.lua
@@ -507,12 +507,9 @@ function ch:preview()
   if fn.has('nvim-0.9') == 1 and ui.title then
     local fname_parts = libs.get_path_info(data[1], 2)
     local tbl = libs.icon_from_devicon(vim.bo[self.main_buf].filetype, true)
-    local theme = require('lspsaga').theme()
     opt.title = {
-      { theme.left, 'TitleSymbol' },
       { (tbl[1] or '') .. ' ', 'TitleFileIcon' },
       { table.concat(fname_parts or {}, libs.path_sep), 'TitleString' },
-      { theme.right, 'TitleSymbol' },
     }
     if #tbl == 2 then
       api.nvim_set_hl(0, 'TitleFileIcon', {
--- a/lua/lspsaga/callhierarchy.lua
+++ b/lua/lspsaga/callhierarchy.lua
@@ -1,6 +1,7 @@
 local api, fn, lsp, uv = vim.api, vim.fn, vim.lsp, vim.loop
 local config = require('lspsaga').config
 local libs = require('lspsaga.libs')
+local window = require('lspsaga.window')
 local call_conf, ui = config.callhierarchy, config.ui
 local insert = table.insert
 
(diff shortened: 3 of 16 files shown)


glepnir:
image

glepnir:
any ui color relate question can discuss there.

glepnir:
image


image

leaxoy:
image

image

After changing colorscheme, highlight broken, so I fix it by adding autocmd, but it should as builtin ability by design.
```lua
vim.api.nvim_create_autocmd("ColorScheme", {
  callback = function() require("lspsaga.highlight").init_highlight() end,
})
```

glepnir:
vote If you don't like currently highlight and color behavior click emoji 😡 .(same buffer buffer highlight  and border missed bug had fixed). I want do some change in highlight and color . because of the recent issue  are full of problems with color highlight. there will have some breakchane I want to do . Can cause trouble for someone just getting used to the current behavior. I'm very sorry . 

thinking:
- follow theme . it will get the color from theme ,if the color is missing fallback the built in color?
- remove all colors. provide many highlight groups ?

glepnir:
new behavior. 

rename_project

glepnir:
now don't need config the highlight

glepnir:
all done. does anyone can give a feedback ?

README.md
```diff
@@ -514,29 +533,13 @@ Default UI options
...
-      yellow = "#f7bb3b",
-      green = "#afd700",
-      cyan = "#36d0e0",
-      blue = "#61afef",
-      purple = "#CBA6F7",
-      white = "#d1d4cf",
-      black = "#1c1c19",
-    },
+    hover = ' ',
     kind = {},
```
nullchilly, line 537: So currently catppuccin's lspsaga kind integration looks like this: https://github.com/catppuccin/nvim/blob/main/lua/catppuccin/groups/integrations/lsp_saga.lua#L21

Did `kind` become a single array with only kinds, I can't find document about this.

glepnir: it just remove colors in config. these colors still used for kind and default highlight groups.

nullchilly: Lspsaga's winbar always gave white background after `ColorScheme` event, did this PR fix the issue?

nullchilly: No it doesn't (Tried out this PR)

Before changing colorscheme

image

After changing colorscheme 

image

glepnir: when you load new colorscheme it will have `hi clear` command to clear all highlight group . I don't think problem should be here ?  unless you defined all winbar highlight group in all of your colorschems or  callback like the comment 
do 

```lua
vim.api.nvim_create_autocmd("ColorScheme", {
  callback = function() require("lspsaga.highlight").init_highlight() end,
})
```

nullchilly: Other plugins have these autocmd on by default.

nullchilly: image

nullchilly: I would opt for removing the colors in `kind` config all together since they do not work if I change to other colorschemes (It keeps the old color table)

Users who want to change highlights can do so through `LspsagaWinbarProperty` if they wish

Otherwise you can support a function like so

```lua
kind = function()
return
end
```

and kind is auto-reloaded on `ColorScheme` event which is precisely how bufferline.nvim currently works

glepnir: I don't think this should plugin do.  support by other way.

nullchilly: Pardon me, did you mean "I don't think this plugin should support the other way."?

nullchilly: Seems to work 👍🏻

nullchilly: Files didn't get redraw...

nullchilly: https://github.com/glepnir/lspsaga.nvim/blob/35815b8aa8d733741130389d91d929ffe9c05223/lua/lspsaga/symbolwinbar.lua#L59

glepnir: which files ? it just redraw winbar highlight. files colorscheme not the lspsaga control

nullchilly: Through `LspSagaWinbarFileIcon` highlight

nullchilly: I will be more clear

glepnir: when you move the cursor it not work ?

glepnir: when you move the cursor it not work ?

glepnir: it taka color from devicons if there is no value it no work.

nullchilly: Here is how it should look with catppuccin-latte
image

Here is how it look after calling `:colorscheme catppuccin-latte` (Previous colorscheme is catppuccin-mocha)

image

nullchilly: Notice that `LspSagaWinbarFileIcon` stays the same after changing from dark to light mode

glepnir: when you move cursor it will restore i think

nullchilly: So it's because `kind` is an array and not a function so it did not get reloaded after 

```lua
kind = require('catppuccin.groups.integrations.lsp_saga').custom_kind(),
```

Which is the reason why I suggest you support kind as a function so it can be reloaded after changing colorscheme

glepnir: nope because it have `default = true` attar.

nullchilly: > when you move cursor it will restore i think

It does not because the user config table is fixed at startup, instead I suggested that you support function type for kind:

```lua
kind = function()
  require('catppuccin.groups.integrations.lsp_saga').custom_kind()
end
```

So that it can be reloaded after `ColorScheme` autocmd

glepnir: because it's a colorscheme  plugin.

glepnir: bug not here

nullchilly: How else do I do it that the colors get reloaded?

glepnir: i missed  a place need a patch .

nullchilly: I don't want to tell users that lspsaga's `kind` array is always fixed so it will be broken if you change colorscheme's flavour

glepnir: fixed.

nullchilly: Since you are against the idea of setting ColorScheme events and I think these patches (nvim_hl_ns_fast) are just workarounds

I just wish I can handle these colors through highlight groups, not through some special arrays

nullchilly: It still doesn't work (as a colorscheme plugin author)

```
kind = require('catppuccin.groups.integrations.lsp_saga').custom_kind()
```

Which is equals to

```
:lua=require('catppuccin.groups.integrations.lsp_saga').custom_kind()                            
{                                                                                                
  Array = { " ", "#89B4FA" },                                                                   
  Boolean = { " ", "#FAB387" },                                                                 
  Class = { " ", "#CBA6F7" },                                                                   
  Constant = { " ", "#89DCEB" },                                                                
  Constructor = { " ", "#89B4FA" },                                                             
  Enum = { "了", "#A6E3A1" },                                                                    
  EnumMember = { " ", "#A6E3A1" },                                                              
  Event = { " ", "#CBA6F7" },                                                                   
  Field = { " ", "#94E2D5" },                                                                   
  File = { " ", "#CDD6F4" },                                                                    
  Function = { " ", "#CBA6F7" },                                                                
  Interface = { " ", "#FAB387" },                                                               
  Key = { " ", "#F38BA8" },                                                                     
  Macro = { " ", "#F38BA8" },                                                                   
  Method = { " ", "#CBA6F7" },                                                                  
  Module = { " ", "#89B4FA" },                                                                  
  Namespace = { " ", "#FAB387" },                                                               
  Null = { " ", "#F38BA8" },                                                                    
  Number = { " ", "#A6E3A1" },                                                                  
  Object = { " ", "#FAB387" },                                                                  
  Operator = { " ", "#A6E3A1" },                                                                
  Package = { " ", "#CBA6F7" },                                                                 
  Parameter = { " ", "#89B4FA" },                                                               
  Property = { " ", "#89DCEB" },                                                                
  StaticMethod = { "ﴂ ", "#FAB387" },                                                            
  String = { " ", "#A6E3A1" },                                                                  
  Struct = { " ", "#CBA6F7" },                                                                  
  TypeAlias = { " ", "#A6E3A1" },                                                               
  TypeParameter = { " ", "#A6E3A1" },                                                           
  Variable = { " ", "#89B4FA" }                                                                 
}
```

Now you see where I am going, the colors are fixed forever, so if I want to switch to a light theme for example the colors remain the same

glepnir: you don't need both have `default` attar if your colorscheme defined it will use them . and your colorscheme override the lspsaga kind color. then  it will use these color even you changed color ? what do you want ? i think the redraw issue had fixed ?

nullchilly: Aren't kind colors **supposed** to get overriden?

glepnir: hmm I don't think i got your point ? you mean when you change the colorscheme like dark to light the color also changed right ?

nullchilly: It is supposed to change the `kind` array too because it has changed from dark to light

nullchilly: But because the kind array is fixed at startup you can't really override it

nullchilly: [Screencast from 2023-01-28 18-03-23.webm](https://user-images.githubusercontent.com/56817415/215263140-fb6c9627-932a-46d3-b2b6-15a64ddbf19b.webm)

glepnir: should be works now.

nullchilly: Still need `kind` to support a function

nullchilly: Ideally please remove the colors from `kind`, this is very problematic with monkeypatched code to redraw

If an user want to override the colors they will just do it through these highlights

`LspsagaWinbarProperty`

`LspsagaWinbarClass`

`LspsagaWinbarStruct`

glepnir: this should work and can be add into callback
```lua
package.loaded['lspsaga.lspkind'] = nil
require('lspsaga.kind').get_kind
```

nullchilly: This will also helps removing lspsaga.nvim from special integration and the highlights will "Just works"

nullchilly: > this should work and can be add into callback
> 
> ```lua
> package.loaded['lspsaga.lspkind'] = nil
> require('lspsaga.kind').get_kind
> ```

It doesn't matter because currently the only way for a colorscheme to set winbar highlights is via:

```lua
require('lspsaga').setup {
  ui = {
    kind = require('catppuccin.groups.integrations.lsp_saga').custom_kind(),
  }
}
```

As you can see the setup table is fixed FOREVER, doesn't matter if you reload the module or not, user config stayed the same

And as fair as I know this is the only way a colorscheme can integrate into lspsaga's winbar which is a little suboptimal.

nullchilly:
```lua
require('lspsaga').setup {
  kind = function()
    require(vim.g.colors_name).palette()
  end
}
```

But if you change the kind to a function however, you can reload it even after setup and it will reload accordingly

nullchilly: In conclusion, [nvim-navic](https://github.com/SmiteshP/nvim-navic/#-customise)'s style works the best here.

glepnir: why i don't public it because I don't like these string in my memory .I only want them run once. all of my plugin will clean the memory after the command or shotcut use.  i like a clean and lightweight things.

glepnir: this way ignored the your feature dynamic changed.  so i need find a better way to support it .

nullchilly: Clean and lightweight thing in this case would be to remove all the colors in kind config and let user set them through neovim's native highlights

I don't want to leave a note like `keep in mind that symbolwinbar won't be able to reload colors`

glepnir: so need more smart way to do it .

nullchilly: So what's wrong with

```lua
vim.api.nvim_set_hl(0, "LspsagaWinbarFile",          {default = true, bg = "#000000", fg = "#ffffff"})
vim.api.nvim_set_hl(0, "LspsagaWinbarModule",        {default = true, bg = "#000000", fg = "#ffffff"})
vim.api.nvim_set_hl(0, "LspsagaWinbarNamespace",     {default = true, bg = "#000000", fg = "#ffffff"})
vim.api.nvim_set_hl(0, "LspsagaWinbarPackage",       {default = true, bg = "#000000", fg = "#ffffff"})
vim.api.nvim_set_hl(0, "LspsagaWinbarClass",         {default = true, bg = "#000000", fg = "#ffffff"})
vim.api.nvim_set_hl(0, "LspsagaWinbarMethod",        {default = true, bg = "#000000", fg = "#ffffff"})
vim.api.nvim_set_hl(0, "LspsagaWinbarProperty",      {default = true, bg = "#000000", fg = "#ffffff"})
vim.api.nvim_set_hl(0, "LspsagaWinbarField",         {default = true, bg = "#000000", fg = "#ffffff"})
vim.api.nvim_set_hl(0, "LspsagaWinbarConstructor",   {default = true, bg = "#000000", fg = "#ffffff"})
vim.api.nvim_set_hl(0, "LspsagaWinbarEnum",          {default = true, bg = "#000000", fg = "#ffffff"})
vim.api.nvim_set_hl(0, "LspsagaWinbarInterface",     {default = true, bg = "#000000", fg = "#ffffff"})
vim.api.nvim_set_hl(0, "LspsagaWinbarFunction",      {default = true, bg = "#000000", fg = "#ffffff"})
vim.api.nvim_set_hl(0, "LspsagaWinbarVariable",      {default = true, bg = "#000000", fg = "#ffffff"})
vim.api.nvim_set_hl(0, "LspsagaWinbarConstant",      {default = true, bg = "#000000", fg = "#ffffff"})
vim.api.nvim_set_hl(0, "LspsagaWinbarString",        {default = true, bg = "#000000", fg = "#ffffff"})
vim.api.nvim_set_hl(0, "LspsagaWinbarNumber",        {default = true, bg = "#000000", fg = "#ffffff"})
vim.api.nvim_set_hl(0, "LspsagaWinbarBoolean",       {default = true, bg = "#000000", fg = "#ffffff"})
vim.api.nvim_set_hl(0, "LspsagaWinbarArray",         {default = true, bg = "#000000", fg = "#ffffff"})
vim.api.nvim_set_hl(0, "LspsagaWinbarObject",        {default = true, bg = "#000000", fg = "#ffffff"})
vim.api.nvim_set_hl(0, "LspsagaWinbarKey",           {default = true, bg = "#000000", fg = "#ffffff"})
vim.api.nvim_set_hl(0, "LspsagaWinbarNull",          {default = true, bg = "#000000", fg = "#ffffff"})
vim.api.nvim_set_hl(0, "LspsagaWinbarEnumMember",    {default = true, bg = "#000000", fg = "#ffffff"})
vim.api.nvim_set_hl(0, "LspsagaWinbarStruct",        {default = true, bg = "#000000", fg = "#ffffff"})
vim.api.nvim_set_hl(0, "LspsagaWinbarEvent",         {default = true, bg = "#000000", fg = "#ffffff"})
vim.api.nvim_set_hl(0, "LspsagaWinbarOperator",      {default = true, bg = "#000000", fg = "#ffffff"})
vim.api.nvim_set_hl(0, "LspsagaWinbarTypeParameter", {default = true, bg = "#000000", fg = "#ffffff"})
vim.api.nvim_set_hl(0, "LspsagaWinbarText",          {default = true, bg = "#000000", fg = "#ffffff"})
vim.api.nvim_set_hl(0, "LspsagaWinbarSeparator",     {default = true, bg = "#000000", fg = "#ffffff"})
```

nullchilly: A lot less headaches

glepnir: because it will need you write too much these codes. try to  refactor kind now .

nullchilly: Really?

```lua
{
	LspsagaWinbarFile = { fg = C.text },
	LspsagaWinbarModule = { fg = C.blue },
	LspsagaWinbarNamespace = { fg = C.peach },
	LspsagaWinbarPackage = { fg = C.mauve },
	LspsagaWinbarClass = { fg = C.mauve },
	LspsagaWinbarMethod = { fg = C.mauve },
	LspsagaWinbarProperty = { fg = C.sky },
	LspsagaWinbarField = { fg = C.teal },
	LspsagaWinbarConstructor = { fg = C.blue },
	LspsagaWinbarEnum = { fg = C.green },
	LspsagaWinbarInterface = { fg = C.peach },
	LspsagaWinbarFunction = { fg = C.mauve },
	LspsagaWinbarVariable = { fg = C.blue },
	LspsagaWinbarConstant = { fg = C.sky },
	LspsagaWinbarString = { fg = C.green },
	LspsagaWinbarNumber = { fg = C.green },
	LspsagaWinbarBoolean = { fg = C.peach },
	LspsagaWinbarArray = { fg = C.blue },
	LspsagaWinbarObject = { fg = C.peach },
	LspsagaWinbarKey = { fg = C.red },
	LspsagaWinbarNull = { fg = C.red },
	LspsagaWinbarEnumMember = { fg = C.green },
	LspsagaWinbarStruct = { fg = C.mauve },
	LspsagaWinbarEvent = { fg = C.mauve },
	LspsagaWinbarOperator = { fg = C.green },
	LspsagaWinbarTypeParameter = { fg = C.green },
	LspsagaWinbarTypeAlias = { fg = C.green },
	LspsagaWinbarParameter = { fg = C.blue },
	LspsagaWinbarStaticMethod = { fg = C.peach },
	LspsagaWinbarMacro = { fg = C.red },
}
```

nullchilly: Seems like an absolute win for me...

glepnir: you can call this to override in colorscheme 

```lua
require('lspsaga.lspkind').init_or_reload(custom)  custom like the `ui.kind` key is icon value is string or table. 
```

nullchilly: Why `string or table`?

I thought we needed `function` support

nullchilly: Ok now the colors are all white by default

image

glepnir: because I don't add the Colorscheme event. just something still working.

glepnir: and i think you can do it in lsp_saga.lua file 
```lua

callback = function()
  require('lspasga.lspkind).init_or_relate(custom_kind())
end

nullchilly: Honestly if you want to remove user config's table just store `config.ui.kind` function, these `init_or_reload` functions are very confusing

nullchilly: I have to go now, just remember that I am not happy with the current state of `kind` because it attracts a lot of monkey code and whatever you do, don't make it harder for colorschemes to support your plugin just because of "lightweight stuff"

glepnir: > just remember that I am not happy 

Are you really interesting? What am I going to remember? Why should I care about your emotions? Who are you? I thought of supported design patterns. But I'm going to rollback.   I've made simple modifications. You just need to add a callback to your stuff. You've been asking me to make huge changes. Just to support a theme.

> don't make it harder for colorschemes to support your plugin just because of "lightweight stuff"

Why do I need support with your stuff? I never use your theme. Really weird. If you can't speak politely. Go back to school and study. Don't harass me

PR #780, opened 2023-01-27
use cwd when lsp client no root_dir

leaxoy:
If `lsp` started in single file mode, there is no `root_dir`, for example: when edit single config file like `alacritty.yaml` with `yamlls` started.

So treat `cwd` as `root_dir` if `bufname` has prefix `root_dir` would be more suitable. Otherwise use absolute path as filename.

### DIFF
--- a/lua/lspsaga/symbolwinbar.lua
+++ b/lua/lspsaga/symbolwinbar.lua
@@ -28,11 +28,14 @@ local function respect_lsp_root(buf)
   if #clients == 0 then
     return
   end
-  local root_dir = clients[1].config.root_dir
+  local root_dir = clients[1].config.root_dir or vim.fn.getcwd()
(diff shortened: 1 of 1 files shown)


lua/lspsaga/symbolwinbar.lua
```diff
@@ -28,11 +28,14 @@ local function respect_lsp_root(buf)
   if #clients == 0 then
     return
   end
-  local root_dir = clients[1].config.root_dir
+  local root_dir = clients[1].config.root_dir or vim.fn.getcwd()
```
glepnir, line 31: `uv.cwd`

glepnir: if this pr works for `single_file_mode` the `root_dir` in there is nil then do some check

Issue #26, opened 2021-01-26
Open folds in previews


weilbith:
Just installed it yesterday, but I'm already enjoying this plugin so much. :blush: 
One thing I realized which is a problem while working is that previews (especially the definition preview) is not open folds. I assume this is just opening the file into a regular buffer and scroll to the specific line. So this automatically applies my usual fold settings (which I use from Treesitter). While I still like my folds and think they are useful when opening a file, for  a preview they are more obstructive. Do you think it is possible to open them? Maybe set the `foldlevel` option for the preview window to a very high value or just "run" `zR` in the window. What ever fits better the code I would say.

---

Another idea: If I remember the LSP specification correctly, the `Location` object includes a range. This would allow to trim the preview to just the range of the definition. So there is nothing shown below. Trimming the buffer seems to cumbersome, but adopting the window height to the range height should be perfectly fine. 
Maybe you are doing this already and I just don't see this as I have the folds. Just came into my mind and I haven't checked the code.

weilbith:
Btw: I'm totally fine to contribute to the plugin myself. I'm just opening issues first to discuss them with you. All the issues must be a lot of work.

glepnir:
Maybe you are right, because I did not test the folding situation, Pr welcome.

Regarding Location, built-in lsp use it everywhere, but I don’t like to use location , so I use a lot of floating windows. I hope our eyes only need to focus on the center of the screen or a slightly lower position instead of lowering our heads. to look at location...

glepnir:
I'm not sure  I understand what you mean. Because my english is not so good :)

weilbith:
> Maybe you are right, because I did not test the folding situation, Pr welcome.

Alright. I'll probably go for the window option approach. Let's see how it goes.

weilbith:
> I'm not sure I understand what you mean. Because my english is not so good :)

I need to express myself better. :grimacing: 

> Regarding Location, built-in lsp use it everywhere, but I don’t like to use location , so I use a lot of floating windows. I hope our eyes only need to focus on the center of the screen or a slightly lower position instead of lowering our heads. to look at location...

I think there is a little misunderstanding. I like the floating window. I mean thats what this plugin is about, right?
[...]

PR #787, opened 2023-01-28
fix floats opens folded

tobealive:
This PR sets the `foldlevel` to the highest fold number of markdown fold headings when creating a floating window.

This should fix an interference when e.g., using https://github.com/preservim/vim-markdown with heading auto folding enabled.

Experienced issue: In my case, I have set headings above level 4 to auto-fold. Normally this should not trigger folding of content for lspsaga, but still floating windows of lspsaga were opened with folded content.

Besides the mentioned plugin it will in general prevent content being opened folded in sagas floating windows.

It was already fixed once before the lspsaga 2.3 refactor in #26 . But setting the `foldlevel` when creating the window got lost along the way.

### DIFF
--- a/lua/lspsaga/codeaction.lua
+++ b/lua/lspsaga/codeaction.lua
@@ -60,6 +60,7 @@ function act:action_callback()
   self.action_bufnr, self.action_winid = window.create_win_with_border(content_opts, opt)
   vim.wo[self.action_winid].conceallevel = 2
   vim.wo[self.action_winid].concealcursor = 'niv'
+  vim.wo[self.winid].foldlevel = 6
   -- initial position in code action window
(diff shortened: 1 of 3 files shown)


glepnir:
Oops I think it should config in your config not lspaga.!!

glepnir:
this will cause bug error. try set `foldlevelstart=999`

tobealive:
I'll give it a shot, thanks. Sorry about the inconvenience.

tobealive:
if one wants to have it specified in the personal config for floating md windows (like the ones from lspsaga):

```lua
au("WinNew", {
	callback = function(win)
		if vim.fn.win_gettype(win.id) == "popup" and vim.bo.filetype == "markdown" then vim.o.foldlevelstart = 99 end
	end,
})
```

PR #797, opened 2023-01-30
Add support for `LspSagaWinbarFileIcon`

### LINKED ISSUE #766: Winbar background incosistency when displaying symbols
### Describe the bug

Not sure if this is really a bug coming from lspsaga, but sometimes when the cursor is in the root of the file (not inside a function or anything), the winbar background becomes lighter, then when entering into some function or any symbol, the background changes to the theme background.

https://user-images.githubusercontent.com/45423661/214205303-8fce9016-7ab5-44a7-9a4b-46afe13d7bd9.mov

I can see this is because when displaying the file root + symbols the winbar content width is 100% of the available space, but when is just the file root, it is not.

Maybe a fix would be to always fit the 100% of the winbar width, or nvim available space.

### Steps to reproduce

using theme: nightfly
not sure if is happening for others themes

### Expected behavior

See always the same background color in the winbar no matter if the cursor is in the file root or inside a element/symbol, etc.

### Neovim version (nvim -v)

0.8.2

### lspsaga commit

latest

### Terminal name/version

iterm2 3.4.19

Racle:
Fixes #766

Ex. `:hi LspSagaWinbarFileIcon guisp=#665c54 gui=underline guibg=#313131`

### DIFF
--- a/lua/lspsaga/libs.lua
+++ b/lua/lspsaga/libs.lua
@@ -34,7 +34,7 @@ function libs.icon_from_devicon(ft, color)
   local icon, hl = libs.devicons.get_icon_by_filetype(ft)
   if color then
     local _, rgb = libs.devicons.get_icon_color_by_filetype(ft)
-    return { icon, rgb }
+    return { icon, rgb, hl }
   end
   return { icon, hl }
 end
@@ -341,6 +341,35 @@ function libs.jump_beacon(bufpos, width)
       end
     end)
   )
+end
+
+function libs.get_hl_by_name(name)
+  local hl = {}
+  if vim.fn.hlexists(name) == 1 then
+    hl = vim.api.nvim_get_hl_by_name("LspSagaWinbarFileIcon", true)
+  end
+
+ return {
+    fg = hl["foreground"],
+    bg = hl["background"],
+    sp = hl["special"],
+    blend = hl["blend"],
+    bold = hl["bold"],
+    standout = hl["standout"],
+    underline = hl["underline"],
+    undercurl = hl["undercurl"],
+    underdouble = hl["underdouble"],
+    underdotted = hl["underdotted"],
+    underdashed = hl["underdashed"],
+    strikethrough = hl["strikethrough"],
+    italic = hl["italic"],
+    reverse = hl["reverse"],
+    nocombine = hl["nocombine"],
+    link = hl["link"],
+    ctermfg = hl["ctermfg"],
+    ctermbg = hl["ctermbg"],
+    cterm = hl["cterm"],
+  }
 end
 
 return libs
--- a/lua/lspsaga/symbolwinbar.lua
+++ b/lua/lspsaga/symbolwinbar.lua
@@ -52,13 +52,18 @@ local function bar_file_name(buf)
   if not res or #res == 0 then
     return
   end
-  local data = libs.icon_from_devicon(vim.bo[buf].filetype)
+  local data = libs.icon_from_devicon(vim.bo[buf].filetype, true)
+
   local bar = bar_prefix()
   local items = {}
   for i, v in pairs(res) do
     if i == #res then
       if #data ~= 0 then
-        table.insert(items, '%#' .. data[2] .. '#' .. data[1] .. ' ' .. '%*')
+        table.insert(items, bar.prefix .. 'FileIcon#' .. data[1] .. ' ' .. '%*')
+        -- set correct devicon color with user set highlights
+        local hl = libs.get_hl_by_name("LspSagaWinbarFileIcon")
+        hl["fg"] = hl["fg"] or data[2]
+        api.nvim_set_hl(0, "LspSagaWinbarFileIcon", hl)
       end
       table.insert(items, bar.prefix .. 'File#' .. v .. '%*')
     else


glepnir:
hmm just add the remove code is enough

Racle:
> hmm just add the remove code is enough

this also fixes issue with wrong icon color (devicons fg was not respected in older version).
In old version icon would be white, but this PR uses devicons colors.

Old version (ex. commit e269e56a142370596261bf85485aa56b89bfe044):
image

This PR:
image

glepnir:
it works fine for me icon color always

HendrikPetertje:
sorry for bumping into this issue :p

Screenshot 2023-01-30 at 16 56 58

Agree with @glepnir on this one. looks good for me too and I'm doing [exotic "lspsaga -> lualine" things](https://git.hendrikpeter.net/hendrikpeter/pico-vim/-/blob/main/lua/peva/plugins/lines.lua#L12)  with all of that successfully too.

I did have problems like this a while ago when i experimented with custom icons (long ago though, before the big overhaul). Do you have a custom icons or color-configuration in your lspsaga settings or would you be willing to share your neovim dotfiles so i can try it out?

Racle:
@HendrikPetertje here's very dirty minimial config (rename to init.lua):
[init.txt](https://github.com/glepnir/lspsaga.nvim/files/10538166/init.txt)


With `glepnir/lspsaga.nvim`:
image


With `racle/lspsaga.nvim`:
image

Basically I'm changing background to little lighter than what lspsaga defaults and I'm using Kitty (terminal emulator) functionality to add underline to winbar so it has little line between code and lspsaga (it's easier to see where lspsaga winbar ends and code starts)

I also tried to change `DevIconLua` directly in lspsaga (while creating this PR) and sadly it changes every `Lua` icon after that (in my "tab" bar and other plugins), so that wasn't a valid choice.

And `DevIcon*` icons are only ones broken (background color doesn't apply and it doesn't have underline with default settings)

glepnir:
```lua
local ok, group = pcall(api.nvim_get_hl_by_name , 'LspsagaWinbarFileIcon')
if ok then
    api.nvim_set_hl(0, "LspsagaWinbarFIleIcon", { fg = group.fg, bg= your bg})
end

````

**Comment by Racle** [comment]
@glepnir

Should this work on new version?

When using `hi LspSagaWinbarFileIcon` icon is white, not real color, but underline works.

When using config within your last message hihglight doesn't change as `LspSagaWinbarFileIcon` is not set

And using my proviced `init.lua` with latest version

If I have this:
```
vim.cmd [[hi LspSagaWinbarFileIcon guisp=#665c54 gui=underline guibg=#313131]]
```
Result is this:
image



If I have this:
```
local ok, group = pcall(vim.api.nvim_get_hl_by_name, "LspsagaWinbarFileIcon")
if ok then
  vim.api.nvim_set_hl(0, "LspsagaWinbarFIleIcon", {fg = group.fg, bg = "#313131"})
end
```

Result is this:
image

**Comment by glepnir** [comment]
need run after winbar rendered.

**Comment by Racle** [comment]
> need run after winbar rendered.

even after scheduling that 2 seconds after starting neovim, it doesn't apply anything

```
local timer = vim.loop.new_timer()
timer:start(
  2000,
  0,
  vim.schedule_wrap(
    function()
      local ok, group = pcall(vim.api.nvim_get_hl_by_name, "LspsagaWinbarFileIcon")
      if ok then
        vim.api.nvim_set_hl(0, "LspsagaWinbarFIleIcon", {fg = group.fg, bg = "#313131"})
      end
    end
  )
)

```

**Comment by glepnir** [comment]
i add  a patch for this  you can just defined background color `Lspsagawinbarfileicon bg =color`

**Comment by Racle** [comment]
But I found workaround, if I create autocmd to run and change colors when buffer loads, everything works as intented.

```
vim.api.nvim_create_autocmd(
  "BufEnter",
  {
    pattern = "*",
    group = vim.api.nvim_create_augroup("__env", {clear = true}),
    callback = function()
      local devicons = require("nvim-web-devicons")
      local _, rgb = devicons.get_icon_color_by_filetype(vim.bo.filetype)
      vim.api.nvim_set_hl(0, "LspsagaWinbarFIleIcon", {fg = rgb, bg = "#313131", special = "#665c54", underline = true})
    end
  }
)
```

**Comment by glepnir** [comment]
please test now just defined the background color.

**Comment by Racle** [comment]
@glepnir works, but requires that change to every buffer.

Ex open neovim with `nvim test.lua` and change `hi`, then run `:e .bash_profile` and you see this:
image

So that would require autocmd to run on BufEnter :thinking: 

(and I still need to figure out how to setup this after winbar. Delaying it by ~600 ms seems to be working, but it's not ideal..)

**Comment by glepnir** [comment]
sorry a mistake bug in that path..

**Comment by Racle** [comment]
@glepnir any change to add underline support to that also?

Background color seems to be working but underline doesn't work (and with autocmd it's blinking when moving cursor etc.).


This should fix underline issue

https://github.com/glepnir/lspsaga.nvim/blob/990be59c4ec904ff5c7d980cb0d067dfa806848e/lua/lspsaga/symbolwinbar.lua#L63
```
        local ok, conf = pcall(api.nvim_get_hl_by_name, 'LspSagaWinbarFileIcon', true)
        local bg
        local underline
        local special
        if ok then
          bg = conf.background
          underline = conf.underline
          special = conf.special
        end
        api.nvim_set_hl(0, 'LspSagaWinbarFileIcon', {
          foreground = data[2],
          bg = bg,
          underline = underline,
          special = special
        })
      end
```

**Comment by glepnir** [comment]
unpack then vim.list_extend  will better.

**Comment by Racle** [comment]
Yep, you're right. This seems to be much more better approach :+1: 

```
        local hl = {foreground = data[2]}
        local ok, conf = pcall(api.nvim_get_hl_by_name, "LspSagaWinbarFileIcon", true)
        if ok then
          vim.list_extend(hl, conf )
        end

        api.nvim_set_hl(0, "LspSagaWinbarFileIcon", hl)
```

**Comment by glepnir** [comment]
I will fix it later. 0.2.5 will publish then

**Comment by Racle** [comment]
@glepnir awesome, thanks!

**Comment by Racle** [comment]
Just noticed, using `vim.list_extend` breaks underline and icon color when moving cursor or changing buffers.

Fix in this https://github.com/glepnir/lspsaga.nvim/pull/797#issuecomment-1409909159 seems to be working just fine.

**Comment by Racle** [comment]
Can confirm, 2.5.0 works flawlessly with this :tada: 

(probably not most efficient way to get all `LspSagaWinbar*` highlights, but it works :shrug:)

```lua
local function set_hl(name)
  -- get lspsaga highlights to list
  local highlights = vim.api.nvim_exec("filter " .. name .. " highlight", true)

  -- loop highlight
  for s in highlights:gmatch("[^\r\n]+") do
    -- split by first space
    local space = s:find(" ") or (#s + 1)
    local string = "hi " .. s:sub(1, space - 1) .. " guisp=#665c54 gui=underline guibg=#313131"
    vim.cmd(string)
  end
end

-- custom winbar colors
vim.cmd [[hi WinBar guisp=#665c54 gui=underline guibg=#313131]]
vim.cmd [[hi WinBarNC guisp=#665c54 gui=underline guibg=#313131]]
vim.cmd [[hi LspSagaWinbarFileIcon guisp=#665c54 gui=underline guibg=#313131]]
set_hl("LspSagaWinbar")
```

glepnir:
```lua
local kind. = require('lspsaga.kind').get_kind()
for _, v in pairs(kind) do
      api.nvim_set_hl(0, 'LspsagaWInbar' .. v[2] , { fg = , bg })
end
```

```

Racle:
Once again, much cleaner solution. Thanks!

Misses few manually set values (LspSagaWinbarSep, LspSagaWinbarWord, LspSagaWinbarSFolderName) but for kinds it works flawlessly :+1:

PR #806, opened 2023-01-31
breaking change: finder and outline

YuCao16:
This PR has two contributors,
1. (breaking change) Add handler to control the height of finder. This update change finder's config table, move key binding to a new key table like other sections.
2. Add config to control enter outline with/without focus.

### DIFF
--- a/lua/lspsaga/finder.lua
+++ b/lua/lspsaga/finder.lua
@@ -334,7 +334,7 @@ function finder:render_finder_result()
     width = window.get_max_content_length(self.contents),
   }
 
-  local max_height = vim.o.lines * 0.5
+  local max_height = math.floor(vim.o.lines * 0.5)
   opt.height = #self.contents > max_height and max_height or #self.contents
   if opt.height <= 0 or not opt.height then
     opt.height = max_height
--- a/lua/lspsaga/finder.lua
+++ b/lua/lspsaga/finder.lua
@@ -334,7 +334,7 @@ function finder:render_finder_result()
     width = window.get_max_content_length(self.contents),
   }
 
-  local max_height = math.floor(vim.o.lines * 0.5)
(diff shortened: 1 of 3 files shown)


glepnir:
just wonder the situation that you want make outline not focus. avoid the wincmd jump into it when you have multiple window right ?

glepnir:
just wonder the situation that you want make outline not focus. avoid the wincmd jump into it when you have multiple window right ?

YuCao16:
> just wonder the situation that you want make outline not focus. avoid the wincmd jump into it when you have multiple window right ?

Yes, sort of. Not all the time I open outline is to go to some location, sometims just want to check the file structure.

YuCao16:
What do you think about the max_height config, any idea to avoid breaking change?

glepnir:
tha is fine I think but in next version will  better. only preview no meanless .support edit would better . so I am refactor preview win now. so I will add this into this version

glepnir:
finder improve in #810  . allowed edit in preview window. can you test the new finder ? and currently I disabled lsp in preview window. in huge project this will load many buffers  may spawn multiple servers. so I disabled it by default.

YuCao16:
> finder improve in #810 . allow edit in preview window. can you test the new finder ? and currently I disabled lsp in preview window. in huge project this will load many buffers may spawn multiple servers. so I disabled it by default.

You may not be able to reproduce, but here is a bug.

[Screencast from 01-02-23 09:38:48.webm](https://user-images.githubusercontent.com/62466929/216006566-bd183441-131b-4403-9015-43b27f481c54.webm)

In addition, after entering finder, cursor is not at the top of the window.

glepnir:
> In addition, after entering finder, cursor is not at the top of the window.

what do you mean in it first definition line ? you video works fine.  only bug is preview window not disappear

PR #824, opened 2023-02-03
fix: add default value when sign text is nil

Avimitin:
I don't use text for sign column to keep the number column neat and tidy. But that will cause lspsaga panic when I invoke the cursor diagnostic action.

This PR check `sign.text` is not nil before getting its length. When the `sign.text` variable is nil, it will default use `1` as col_end value.


Avimitin:
Lua format fixed.

PR #825, opened 2023-02-03
Option to show name of the LSP in code actions

MaximilianLloyd:
This is useful if you have many active LSPs. Won't be useful for everyone though which is why i've set it to be disabled by default.

### DIFF
--- a/doc/lspsaga.nvim.txt
+++ b/doc/lspsaga.nvim.txt
@@ -301,6 +301,7 @@ Default options:
 >
       code_action = {
         num_shortcut = true,
+	show_server_name = false,
         keys = {
           -- string | table type
           quit = "q",
--- a/doc/lspsaga.nvim.txt
+++ b/doc/lspsaga.nvim.txt
@@ -301,7 +301,6 @@ Default options:
 >
       code_action = {
         num_shortcut = true,
-	show_server_name = false,
         keys = {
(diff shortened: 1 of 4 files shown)


doc/lspsaga.nvim.txt
```diff
@@ -301,6 +301,7 @@ Default options:
 >
       code_action = {
         num_shortcut = true,
+	show_server_name = false,
```
glepnir, line 304: don't edit this file . it generate by ci

lua/lspsaga/codeaction.lua
```diff
@@ -148,8 +151,9 @@ function act:send_code_action_request(main_buf, options, cb)
     self.action_tuples = {}
 
     for client_id, result in pairs(results) do
+			local name = vim.lsp.get_client_by_id(client_id).name
```
glepnir, line 154: these code in wrong pos. you should get the name when the option equals to true.

glepnir (changes requested):
logic is not correct. tuples still 2 . just get client name in option check then set to title.

MaximilianLloyd:
> 

Good point :smile: . Fixed now.

MaximilianLloyd:
Sorry about the wrong formatting. I ran stylua now to fix it.

PR #841, opened 2023-02-05
fix: error when showing diagnostic if lsp signs are disabled

protiumx:
When LSP signs are disabled `vim.diagnostic.config({ signs = false })`, calling `:Lspasaga show_cursor_diagnostics` produces the following error
```
Error executing Lua callback: ...ack/packer/start/lspsaga.nvim/lua/lspsaga/diagnostic.lua:496: attempt to index local 'sign' (a nil value)
```
When checking the source I noticed that when signs are disabled calling `vim.fn.sign_getdefined` returns an empty list.

Note: formatting was applied by my LSP suneko Lua server

### DIFF
--- a/lua/lspsaga/diagnostic.lua
+++ b/lua/lspsaga/diagnostic.lua
@@ -57,7 +57,7 @@ function diag:code_action_cb()
   end
 
   local fix_title = diag_conf.custom_fix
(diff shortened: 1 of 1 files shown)


glepnir:
format with stylua

Issue #850, opened 2023-02-09
Go to type definition


matthewsia98:
Provide `Lspsaga goto_type_definition` command for completeness

loicdiridollou:
@matthewsia98  please check out the pull request I opened regarding this new feature, it works similarly to go to definition and will allow completeness of the match between Lspsaga and lsp standard behavior.

PR #856, opened 2023-02-10
Adding go to type definition functionality

loicdiridollou:
Cf issue #850 

* Adding go to type definition functionality mimicking the go to definition implementation.
* Updating the docs with the relevant information to inform on the feature.

Happy to make any changes in the docs or code, I wasn't able to reproduce the same theme as the other showcase so decided against adding an image for the sake of uniformity of style in the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,17 @@ keymap("n", "gd", "<cmd>Lspsaga peek_definition<CR>")
 -- Go to definition
 keymap("n","gd", "<cmd>Lspsaga goto_definition<CR>")
 
+-- Peek type definition
+-- You can edit the file containing the type definition in the floating window
+-- It also supports open/vsplit/etc operations, do refer to "definition_action_keys"
+-- It also supports tagstack
+-- Use <C-t> to jump back
+keymap("n", "gt", "<cmd>Lspsaga peek_type_definition<CR>")
+
+-- Go to type definition
+keymap("n","gt", "<cmd>Lspsaga goto_type_definition<CR>")
+
+
 -- Show line diagnostics
 -- You can pass argument ++unfocus to
 -- unfocus the show_line_diagnostics floating window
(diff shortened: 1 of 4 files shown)


glepnir:
why not simply do it in definition.lua . this way used lots of duplicate codes

loicdiridollou:
> 

You will need to create a different group for the peek_type_definition with the api.nvim_create_autogroup. Let me work on this and propose the changes, that would indeed avoid duplicating a lot of code, as long as code remains readable for other users that should work.

lua/lspsaga/definition.lua
```diff
@@ -296,6 +296,155 @@ function def:goto_definition()
...
+
+          remove(index)
+        end
+      end,
+    })
+  end)
+end
+
+-- override the default the type defintion handler
+function def:goto_type_definition()
```
glepnir, line 430: these also duplicate codes. can we abstract a public function like `def:goto_definition(method)`

loicdiridollou: Thanks for the review, indeed that will cut down the duplicates even more, i have used `strings.format` to pass the method in the type of handlers required for the method. Let me know if you have a better way to handle the strings, I preferred to pass only the second part of the handler endpoint for simplicity (`definition` instead of `textDocument/definition`)

glepnir: use number will be better  in command paramter it just take 1 byte. a local function which have method table. this table will marked by gc when the invoked function finished.   While numbers as parameters. make the meaning unclear. But it's more convenient.

```lua

--commandl.lua
peek_defintion(1)  1 or 2

-- definition.lua
local function get_method(index)
  local tbl = { method1, method2}
  return tbl[index]
end
```

lua/lspsaga/definition.lua
```diff
@@ -171,11 +176,11 @@ function def:peek_definition()
 
   local params = lsp.util.make_position_params()
 
-  lsp.buf_request_all(current_buf, 'textDocument/definition', params, function(results)
```
glepnir, line 174:
```lua
local method = get_method(index)
```

lua/lspsaga/definition.lua
```diff
@@ -171,11 +176,11 @@ function def:peek_definition()
...
+  lsp.buf_request_all(current_buf, get_method(method), params, function(results)
     in_process = 0
     if not results or next(results) == nil then
       vim.notify(
-        '[Lspsaga] response of request method textDocument/definition is nil',
+        string.format('[Lspsaga] response of request method %s is nil', get_method(method)),
```
glepnir, line 183: there is no need run get_method twice. just concat method in there.

README.md
```diff
@@ -239,6 +250,26 @@ src="https://user-images.githubusercontent.com/41671631/215719806-0dea0248-4a2c-
 
 Jumps to the definition of the hovered word and shows a beacon highlight.
 
+## :Lspsaga peek_type_definition
```
glepnir, line 253: this section is no need. they both use same keymap

PR #867, opened 2023-02-13
version 0.2.7

glepnir:
## Most
Since the ui changed in 0.2.3. I have received a lot of unsatisfactory feedback. Yes. Most people don't like to spend a lot of time configuring the ui on the plugin. Especially in color. So I slowly remove new changes in subsequent versions. Some use the old ui and some use the new behaviour. For example `lspfinder incomingcall outgoing call`. Main window and preview window are combined. And support any normal buffer operation. The current diagnostic jump window is more concise.

`ui.colors` has been completely removed. And the built-in color settings have been completely removed. The color about the kind is set using the highlight group. This can more closely match the current theme. And provide api more convenient configuration. outline uses `winbarkindname` to highlight groups.

## Features
- diagnostic on insert like emacs. helix . `diagnostic.on_insert` option
  default is true. false to disable it. for better usage you should disable the built in diagnsotic virtualtext in insert. and there has commands `DiagnosticInsertEnable DiagnosticInsertDisable` that you can enable or disable it by manually. see pic 1
- `on_insert_follow` default is false . when is false on top right. true follow your line number
- `insert_winblend` default is 0, when it's to 100 will completely transparent. the color will
  changed a little light. 0 will use the `NormalFloat` group. it will link to `Normal` by Lspsaga.
- `symbol_in_winbar.ignore_patterns` when the filename matched the pattern .will ignore render symbols. when the show_file is true it sill render file folder.

## BugFix

- `show diagnostics` can't jump into without unfocus set .

## New apis
- `require('lspsaga.kind').get_kind_group` it will return all the `SagaWinbar`  + `kind name` group . also include `SagaWinbarFileName SagaWinbarFileIcon SagaWinbarFolderName SagaWinbarSep`.these groups is special . so if you want use this api to custom the highlight. you need dealwith these 4 groups the last item is `SagaWinbarSep`

## BreakChange
- `diagnostic`  jump  ui more easy, `show_buf show_line show_cursor ui changed ` add count in head
- `code action` ui back to old.
- `LspSagaWinbar` prefix change to `SagaWinbar`.
-  `LSOutline` highlights removed , use `SagaWinbar`  instead.
-  `custom_kind` the value not colors code, use the highlight group instead.

## Changed
- `codeaction.extend_gitsings` to false in default

## Removed
- `config.diagnostic.custom_fix`
- `config.diagnostic.custom_msg`
- `ui.diagnostic`
- `ui.colors`
- `ui.theme`


glepnir:
min

README.md
```diff
@@ -333,6 +332,8 @@ Jumps to next diagnostic position and show a beacon highlight. Lspsaga will then
 Default options:
 ```lua
   diagnostic = {
+    on_insert = false,
```
Jint-lzxy, line 335:
```suggestion
    on_insert = true,
```
`lua/lspsaga/init.lua` sets this option to true by default. Is this a typo?

README.md
```diff
@@ -359,6 +360,8 @@ Default options:
   text color
 - `border_follow` the border highlight will follow the diagnostic type. if false it will use the
   highlight `DiagnosticBorder`.
+- `on_insert` default is false it works like the emacs helix show diagnostic in right but in line.
```
Jint-lzxy, line 363:
```suggestion
- `on_insert` default is true it works like the emacs helix show diagnostic in right but in line.
```
and this line

PR #904, opened 2023-02-22
Check if any servers support code actions in diagnostics

iovis:
Love the new release!

I stumbled upon the following error jumping to a diagnostic (`:Lspsaga diagnostic_jump_prev`) when none of the LSPs support code actions:

```
method textDocument/codeAction is not supported by any of the servers registered for the current buffer
```

I tracked it to this line:
https://github.com/glepnir/lspsaga.nvim/blob/688a3d835a4533ebd2c1ebb7e7c22d53385975d5/lua/lspsaga/codeaction.lua#L163

Looking at other parts of your code, it looks like it's fallible and you had a check for it. For example, `send_request()` has a similar call:
https://github.com/glepnir/lspsaga.nvim/blob/688a3d835a4533ebd2c1ebb7e7c22d53385975d5/lua/lspsaga/lightbulb.lua#L112

And it has the following check before it gets called:
https://github.com/glepnir/lspsaga.nvim/blob/688a3d835a4533ebd2c1ebb7e7c22d53385975d5/lua/lspsaga/lightbulb.lua#L136-L142

So I extracted `check_server_support_codeaction(bufnr)` to `libs` and added it as a check before asking for the possible code actions

### DIFF
--- a/lua/lspsaga/diagnostic.lua
+++ b/lua/lspsaga/diagnostic.lua
@@ -169,7 +169,10 @@ function diag:render_diagnostic_window(entry, option)
   vim.list_extend(content, convert)
   content[#content] = content[#content] .. source
 
-  if diag_conf.show_code_action then
+  if
+    diag_conf.show_code_action
+    and libs.get_client_by_cap({ 'codeActionProvider', 'resolveProvider' })
+  then
     act:send_code_action_request(self.main_buf, {
       range = {
         start = { entry.lnum + 1, entry.col },
--- a/lua/lspsaga/libs.lua
+++ b/lua/lspsaga/libs.lua
@@ -242,8 +242,7 @@ function libs.get_client_by_cap(caps)
     ['table'] = function(instance)
       libs.add_client_filetypes(instance, { vim.bo.filetype })
       if
-        instance.server_capabilities[caps[1]]
-        and instance.server_capabilities[caps[2]]
+        vim.tbl_get(instance.server_capabilities, unpack(caps))
         and libs.has_value(instance.config.filetypes, vim.bo.filetype)
       then
         return instance


glepnir:
you can use `libs.get_client_by_cap` to get the client. nil non support then.

iovis:
I just tried with your suggestion, but I may be misunderstanding something on the implementation

I tried:
```lua
-- lua/lspsaga/diagnostic.lua
if diag_conf.show_code_action and libs.get_client_by_cap('codeActionProvider') then
```

But I still ran into the error. This is the relevant differences for `server_capabilities` between two LSP servers I'm using (`null-ls` and `rust-analyzer`):
```lua
-- null-ls (failing)
{
  server_capabilities = {
    codeActionProvider = {
      resolveProvider = false  -- not getting checked in `libs.get_client_by_cap`
    },
  },
}
```
```lua
-- rust-analyzer (working)
{
  server_capabilities = {
    codeActionProvider = {
      codeActionKinds = { "", "quickfix", "refactor", "refactor.extract", "refactor.inline", "refactor.rewrite" },
      resolveProvider = true
    },
  },
}
```

I saw you have some explicit checks in `codeaction.lua` for `resolveProvider`:
https://github.com/iovis/lspsaga.nvim/blob/d0192f71eb085a68ba19fcf4f6597a503e03f9e3/lua/lspsaga/codeaction.lua#L277
https://github.com/iovis/lspsaga.nvim/blob/d0192f71eb085a68ba19fcf4f6597a503e03f9e3/lua/lspsaga/codeaction.lua#L303

But I'm unsure how you prefer to solve it

glepnir:
sorry for late. use vim.tbl_get would be great.

iovis:
I updated the code to leverage `libs.get_client_by_cap` and `vim.tbl_get`, let me know if the implementation makes sense to you.

I took the liberty to change the logic a bit for the `["table"]` branch because I didn't see any other parts of the code using it with a table anymore. So instead of checking for the first `[1]` and second `[2]` values of a list, it treats it as something to unpack for `vim.tbl_get()`. If I missed any uses of that function or you were planning to use it that way in the future, let me know and we can think of something else.

PR #917, opened 2023-02-25
fix: close float window when leave buffer

NavePnow:


glepnir:
this is not correct. now it can works when you leave buffer.

NavePnow:
> this is not correct. now it can works when you leave buffer.

when you click buffers to switch them, the float window is not close.

https://user-images.githubusercontent.com/39482599/221352877-89b47e89-983c-49bc-913f-d3770ed96217.mov

glepnir:
use `++unfocus` or use buffer  jump command bnext bprev . i don't use mouse.

NavePnow:
the problem still shows up when adding `++unfocus`.

NavePnow:
is BufLeave event conflict with lspsaga's current settings? since you mentioned it is not correct.

glepnir:
if you used the unfocus if also append the `BufLeave` into it ?  if you don't use unfocus that mean you can jump into that window. it's useful to check diagnostic and jump to pos in show_line and show_buf diangostics. wincmd will also trigger BufLeave the event. please see code carefully i use bufenter instead of this.   if you want make a pr you need look code context.

PR #967, opened 2023-03-26
hover: change fallback browser to xdg-open

sorashi:
https://github.com/glepnir/lspsaga.nvim/blob/d47c2dbfa16b95563233bf23703e9f8b04b7a969/lua/lspsaga/hover.lua#L36-L42

A better fallback URI opener is `xdg-open`, it is the most widespread [resource opener](https://wiki.archlinux.org/title/default_applications#Resource_openers). It also works on [BSD](https://man.freebsd.org/cgi/man.cgi?query=xdg-open&sektion=1&manpath=freebsd-release-ports). Other users will have to change the configuration option to suit their needs.

By the way, note that most Linux distributions don't use `chrome`, but `google-chrome`, or `google-chrome-stable`. A lot of Linux users use `chromium-browser` as it's non-proprietary. Not to mention users of other internet browsers.

### DIFF
--- a/lua/lspsaga/init.lua
+++ b/lua/lspsaga/init.lua
@@ -21,7 +21,7 @@ local default_config = {
   hover = {
     max_width = 0.6,
     open_link = 'gx',
-    open_browser = '!chrome',
+    open_browser = '!xdg-open',
   },
   diagnostic = {
     on_insert = true,


glepnir:
xdg-open only for linux . note not every one in linux. you can config in your local

glepnir:
xdg-open only for linux . note not every one in linux. you can config in your local

sorashi:
`chrome` is a **worse** default, but ok

glepnir:
for more better can we check the system version and give it a default value ? like `uv.os_uname` is linux then `xdg-open google-chrome or firefox` if mac ...if win ...

sorashi:
So you're saying
```
win -> '!start cmd /cstart /b '
mac -> 'silent !open '
linux -> '!xdg-open '
bsd -> '!xdg-open '
other -> '!chrome '
```
Better leave it as it is now.

glepnir:
yep something like this. and can we get the default brower name in lua or c ?

sorashi:
> and can we get the default brower name in lua or c ?
no, because every OS uses a different way to save the default browser. Also some Linux don't have `xdg-open`, it uses something else. Just leave it be.
Closed. :smile:

glepnir:
okay :)

PR #970, opened 2023-03-27
Load lspsaga using `LspAttach` instead of `BufRead`

UnrealApex:


UnrealApex:
I was not aware that Neovim has LSP specific events, thank you for pointing that out!

glepnir:
I fixed .

PR #978, opened 2023-03-28
fix ++unfocus arg

buraksirma:
Hey, it looks like ++unfocus flag is not respected right now.  I've a one minor question out of topic. Will you add `Lspsaga show_cursor_dignostics` back. I was using it a lot and my workflow has broken after the last update here is the my work around for now.

```
lua require('lspsaga.showdiag'):show_diagnostics({ cursor = true, arg = '++unfocus' })
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,10 +122,6 @@ keymap("n","gt", "<cmd>Lspsaga goto_type_definition<CR>")
 -- unfocus the show_line_diagnostics floating window
 keymap("n", "<leader>sl", "<cmd>Lspsaga show_line_diagnostics<CR>")
 
--- Show cursor diagnostics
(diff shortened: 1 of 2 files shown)


glepnir:
> Will you add Lspsaga show_cursor_dignostics

did you used diagnsotic jump this will also show the cursor diagnostic first. and show actions in this window. more useful than before. and `opt.arg` is not correct . opt is `nvim_open_win` params.

glepnir:
back show_cursor is easy .

glepnir:
maybe we can add this new cursor show into the readme ?

glepnir:
I had add it back `Lspsaga  show_cursor_diagnsotics`

buraksirma:
@glepnir Thank you :tada:

PR #995, opened 2023-04-04
fix: any hover client succeeding suppresses 'No information available'

johnpyp:
Currently, if there are two lsp clients that may report supporting `textDocument/hover`, and one of them "fails" the hover while the other succeeds, a hover will popup **and** "No information available" will be notified.

This PR makes it so that "No information available" is only notified if all the possible clients have have also failed.

(Note: I didn't use `lsp.buf_request_all` because that doesn't pass the `ctx` object to the callback, which is currently used)

Example:
image

The two servers that are being activated are `tsserver` and `tailwindcss`. Tailwindcss will only hover in cases where you're hovering over a tailwindcss class, so it makes sense that it returning `nil` for the response.

This PR doesn't address that two clients returning non-null values may overlap/have a race condition - I'm not sure what the expected behavior should be in that situation

### DIFF
--- a/lua/lspsaga/hover.lua
+++ b/lua/lspsaga/hover.lua
@@ -211,14 +211,38 @@ end
 
 function hover:do_request(args)
   local params = util.make_position_params()
+
+  local expected_result_count = 0
+  lsp.for_each_buffer_client(0, function(client)
+    if client.supports_method('textDocument/hover') then
+      expected_result_count = expected_result_count + 1
+    end
+  end)
+
+  local result_count = 0
+  local has_succeeded = false
+
+  local should_error = function()
+    -- Never error if we have ++quiet
+    if args and has_arg(args, '++quiet') then
+      return false
+    end
+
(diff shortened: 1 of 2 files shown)


lua/lspsaga/hover.lua
```diff
@@ -211,14 +211,38 @@ end
 
 function hover:do_request(args)
   local params = util.make_position_params()
+
+  local expected_result_count = 0
+  lsp.for_each_buffer_client(0, function(client)
```
glepnir, line 216: does there can use `libs.get_clinet_by_cap` by using `hoverProvider`

johnpyp: The issue with that is that it would only return the first one that *could* return a valid hover result. In the case of the two servers, if you hover over a string with a tailwindcss class in it, tailwindcss server will return a hover and tsserver won't, but otherwise in most other cases only tsserver will return a hover result. If I used get_client_by_cap, it would choose one of the two for all supported filetypes right?

PR #996, opened 2023-04-04
docs: clarify usage of .setup() options

johnpyp:
Clarify the usage of `.setup` with non-empty options in the readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,6 +183,8 @@ You can find the documentation for Lspsaga in Neovim by using `:h lspsaga`.
 
 ## Default options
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -195,6 +197,30 @@ You can find the documentation for Lspsaga in Neovim by using `:h lspsaga`.
   request_timeout = 2000,
 ```
 
+Example setup using default options:
+
+```lua
+require("lspsaga").setup({
+  preview = {
```
glepnir, line 204: just a few options is enough other use `...` instead ? because it's an example.

Issue #975, opened 2023-03-27
Error on certain entries in call hierarchy


mikehaertl:
### Describe the bug

Some entries in the call hierarchy give an error:

```
Fehler beim Ausführen von "CursorMoved Autokommandos für "<buffer=18>"":
Error executing lua callback: .../nvim/plugged/lspsaga.nvim/lua/lspsaga/callhierarchy.lua:522: Cursor position outside buffer
stack traceback:
        [C]: in function 'nvim_win_set_cursor'
        .../nvim/plugged/lspsaga.nvim/lua/lspsaga/callhierarchy.lua:522: in function 'preview'
        .../nvim/plugged/lspsaga.nvim/lua/lspsaga/callhierarchy.lua:380: in function <.../nvim/plugged/lspsaga.nvim/lua/lspsaga/callhierarchy.lua:379>
```
It happens when I try to move the cursor over the last entry `IllegalStateException` in the list here:

![image](https://user-images.githubusercontent.com/675062/228000427-52b641ae-3c73-4800-9811-3dfdadc69aec.png)

### Steps to reproduce

```java
    public @NonNull RenderResult render(@NonNull RenderJob job) throws Exception {
        final EmailDef def = job.getEmailDef();
        final Content content = def.getContent();
        switch (content.getType()) {
            case DIRECT:
                return renderDirectEmail(job);
            case TEMPLATE:
                return renderTemplateEmail(job);
            case NAMED:
                return renderNamedEmail(job);
            default:
                throw new IllegalStateException("Unhandled email definition content type: " + content.getType());
        }
    }
```

* `:Lspsaga outgoing_calls`
* Move cursor over `IllegalStateException`

### Expected behavior

no error

### Neovim version (nvim -v)

0.8.3

### lspsaga commit

3ae79a1

### Terminal name/version

Gnome Terminal 3.44.0

glepnir:
these code not work on my side ..I don't write java . can you provide a complete min java code ..

mikehaertl:
Also related:
https://github.com/mfussenegger/nvim-jdtls/blob/f8fb45e05e638e5c67e884f3039abcda7abc2d2d/plugin/nvim_jdtls.vim#L6

glepnir:
the log not show the response of `outgoing` request method ..

mikehaertl:
Here's the full log when I issue `:Lspsaga outgoing_calls`:

```
[...]

PR #1029, opened 2023-04-18
#975: Fix JDT previews in "outgoing_calls" popup.

blz777:
## Why?
When executing `:Lspsaga outgoing_calls` for a java method, the preview for `jdt` based nodes (e.g. class files outside of the project) does not currently work. 
before-fix

This PR aims to fix the preview of the files using a JDT LS client.

## What?
Use a JDT LS client to display the node content when previewing a specific node. The implementation is based on `nvim-jdtls/lua/jdtls.lua`'s `function M.open_classfile(fname)` method. Since `open_classfile` opens the file in the current buffer, I had to provide a separate method, which uses an already created buffer.  

Ultimately, the `load_jdt_preview` function should be part of the https://github.com/mfussenegger/nvim-jdtls plugin. Once such a method is available there, we can drop the `load_jdt_preview` method.

## Testing Done
1. Opened a java project.
2. Executed `:Lspsaga outgoing_calls`
3. Navigated to CountDownLatch and observed that the preview of the file is displayed correctly and the respective method (the constructor in this case) is visible in the preview.
Screenshot 2023-04-18 at 10 58 54
4. Navigated a few nodes below to the `assertTrue(boolean, String)` entry and made sure the method is visible in the preview window.
 
Screenshot 2023-04-18 at 10 58 59

as expected.

## Notes
Fixes https://github.com/nvimdev/lspsaga.nvim/issues/975.

### DIFF
--- a/lua/lspsaga/callhierarchy.lua
+++ b/lua/lspsaga/callhierarchy.lua
@@ -458,9 +458,9 @@ local function load_jdt_preview(bufnr, uri)
   vim.bo[bufnr].filetype = 'java'
 
   vim.wait(config.request_timeout, function()
-    return next(lsp.get_active_clients({ name = 'jdtls', bufnr = bufnr})) ~= nil
+    return next(lsp.get_active_clients({ name = 'jdtls', bufnr = bufnr })) ~= nil
   end)
-  local client = lsp.get_active_clients({ name = 'jdtls', bufnr = bufnr})[1]
+  local client = lsp.get_active_clients({ name = 'jdtls', bufnr = bufnr })[1]
   assert(client, 'Must have a `jdtls` client to load class file or jdt uri')
 
   local content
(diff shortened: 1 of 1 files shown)


lua/lspsaga/callhierarchy.lua
```diff
@@ -449,6 +449,39 @@ function ch:get_node_at_cursor()
...
 
+local function load_jdt_preview(bufnr, uri)
+    vim.bo[bufnr].modifiable = true
+    vim.bo[bufnr].swapfile = false
+    vim.bo[bufnr].buftype = 'nofile'
+
+    -- This triggers FileType event which should fire up the lsp client if not already running.
+    vim.bo[bufnr].filetype = 'java'
+
+    local timeout_ms = 5000
```
glepnir, line 460: there can use `config.request_timeout` default is 2000.

PR #1041, opened 2023-04-25
bad argument #1 to 'sub' (number expected, got nil)

haringsrob:
I sometimes receive an error: `...site/pack/packer/opt/lspsaga.nvim/lua/lspsaga/finder.lua:277: bad argument #1 to 'sub' (number expected, got nil)`

This seems to fix it


lua/lspsaga/finder.lua
```diff
@@ -273,7 +273,7 @@ function finder:create_finder_data(result, method)
       start_col = node.col - 10
     end
     node.word = api.nvim_buf_get_text(node.bufnr, node.row, start_col, node.row, node.ecol, {})[1]
-    if node.word:find('^%s') then
```
glepnir, line 276: there check the line has indent . so we need check the value of `node.word:find('%S')` is nil or not in sub function

PR #1063, opened 2023-05-12
Show classname of outgoing call in preview.

blz777:
## Why?
When using `Lspsaga outgoing_calls`, the methods being called from a specific method get displayed in the list on the left side. In order to identify which class the respective method belongs to, we have to move the cursor to the line and see the name of the class in the preview window on the right. It would be more convenient if we can see all the target classes that the methods belong to at a glance.

## What?
Add the name of the class that the respective method in the preview belongs to using a ' @ <shortClassName>' suffix.
When the class is the same as the class the method was called, the shortClassName will be empty and the suffix would be ' @'.

## Testing Done
Before change, executed `Lspsaga outgoing_calls`:
Screenshot 2023-05-12 at 9 49 00

After change, executed `Lspsaga outgoing_calls`:
Screenshot 2023-05-12 at 9 50 19

We can see that the class names are present with the expected suffix in all methods being called.

If the method call is to the same class, we can see that the suffix is simply ' @' without any class:
Screenshot 2023-05-12 at 9 58 08

as expected.

### DIFF
--- a/lua/lspsaga/callhierarchy.lua
+++ b/lua/lspsaga/callhierarchy.lua
@@ -122,7 +122,7 @@ function ch:call_hierarchy(item, parent)
         icons[#icons + 1] = kind[target.kind]
         local expand_collapse = '  ' .. ui.expand
         local icon = kind[target.kind][2]
-        local className = target.uri:match(".+/(.+)[.]class[?]")
+        local className = target.uri:match('.+/(.+)[.]class[?]')
         className = className and className or ''
         self.data[#self.data + 1] = {
           target = target,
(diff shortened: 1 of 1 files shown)


blz777:
Hi @glepnir,
When opening Pull Requests, do I also need to open an Issue, or would the PR suffice as it is?

Thank you,
Anton

glepnir:
Nope. LGTM yesterday github crash. so i didn't review it yesterday.

blz777:
Thank you so much for the rapid approval!
I seem to have missed another use case, which I covered here:
https://github.com/nvimdev/lspsaga.nvim/pull/1065

PR #1065, opened 2023-05-12
Fix project class names in outgoing calls preview.

blz777:
## Why?
Some class names are not visible in `Lspsaga outgoing_calls` for methods which reside in classes, referenced in the project itself, i.e. by java source instead of by class file. We need to enhance the capturing of the class name in the preview window. 
The current code handles properly urls, which are:
>jdt://contents/slf4j-api-1.7.30.jar/org.slf4j/Logger.class?=server/%5C/Users%5C/user%5C/.m2%5C/repository%5C/org%5C/slf4j%5C/slf4j-api%5C/1.7.30%5C/slf4j-api-1.7.30.jar=/gradle_used_by_scope=/integrationTest,main,test=/%3Corg.slf4j(Logger.class

but the following urls are not handled properly:
>file:///Users/user/source/atlas/cloud-services-agents/vsphere-inventory-agent/server/src/main/java/com/company/product/cloud/inventory/monitor/MonitorRegistry.java

## What?
Enhance the regular expression for parsing the class name for the outgoing calls preview to capture properly in the case of java source file url.

## Testing Done
Before change:
before
we can see the missing class names in the list of methods.
After change:
after

We can see all class names properly.

as expected.

### DIFF
--- a/lua/lspsaga/callhierarchy.lua
+++ b/lua/lspsaga/callhierarchy.lua
@@ -127,7 +127,6 @@ function ch:call_hierarchy(item, parent)
         local classNamePart = ''
 
         -- Class name resolution for Java
-        -- 
         if vim.bo[api.nvim_get_current_buf()].filetype == 'java' then
           local projectClassPattern = '.+/([^/]+)[.]java'
           local jdtClassPattern = '/([^/?=]+)[.]class'
--- a/lua/lspsaga/callhierarchy.lua
+++ b/lua/lspsaga/callhierarchy.lua
@@ -123,16 +123,25 @@ function ch:call_hierarchy(item, parent)
         local expand_collapse = '  ' .. ui.expand
         local icon = kind[target.kind][2]
 
-        local projectClassPattern = '.+/([^/]+)[.]java'
-        local jdtClassPattern = '/([^/?=]+)[.]class'
-        local projectClass = target.uri:match(projectClassPattern)
-        local jdtClass = target.uri:match(jdtClassPattern)
-        local className = projectClass or jdtClass
-        className = className and className or ''
+        -- Variable can be used to append name of class to the respective method in the outgoing preview popup.
+        local classNamePart = ''
+
+        -- Class name resolution for Java
+        -- 
+        if vim.bo[api.nvim_get_current_buf()].filetype == 'java' then
+          local projectClassPattern = '.+/([^/]+)[.]java'
+          local jdtClassPattern = '/([^/?=]+)[.]class'
+          local projectClass = target.uri:match(projectClassPattern)
+          local jdtClass = target.uri:match(jdtClassPattern)
+          local className = projectClass or jdtClass
+          if className ~= nil then
+            classNamePart = ' @ ' .. (className and className or '')
+          end
+        end
 
         self.data[#self.data + 1] = {
           target = target,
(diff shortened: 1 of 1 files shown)


glepnir:
```
file:///Users/user/source/atlas/cloud-services-agents/vsphere-inventory-agent/server/src/main/java/com/company/product/cloud/inventory/monitor/MonitorRegistry.java
```
this is a normal uri in lsp  .

```
jdt://contents/slf4j-api-1.7.30.jar/org.slf4j/Logger.class?=server/%5C/Users%5C/user%5C/.m2%5C/repository%5C/org%5C/slf4j%5C/slf4j-api%5C/1.7.30%5C/slf4j-api-1.7.30.jar=/gradle_used_by_scope=/integrationTest,main,test=/%3Corg.slf4j(Logger.class
```

this is very special . language server special case makes things a bit troublesome.  I don't know if other language servers return this similar uri if so. We can make a generic uri parser.

lua/lspsaga/callhierarchy.lua
```diff
@@ -122,8 +122,14 @@ function ch:call_hierarchy(item, parent)
         icons[#icons + 1] = kind[target.kind]
         local expand_collapse = '  ' .. ui.expand
         local icon = kind[target.kind][2]
-        local className = target.uri:match('.+/(.+)[.]class[?]')
+
+        local projectClassPattern = '.+/([^/]+)[.]java'
```
glepnir, line 126: if this only for  java check buffer filetype as first ?

blz777: Done - I have extracted the new functionality to be executed only if the filetype of the current buffer is java.
Left the classNamePart variable once we start implementing this for other languages as well.

fwiw, when I try to execute `Lspsaga outgoing_calls` in python or lua file (with LSP working), I get the following:
>method textDocument/prepareCallHierarchy is not supported by any of the servers registered for the current buffer

I will investigate further when I have more time. I have to fix this for my setup, so I can contribute for other languages as well (and handle them better whenever I add some functionality for java).

Thank you!

blz777:
Hi @glepnir,
Do you think we can handle the support for other languages in a separate PR?

blz777:
https://github.com/nvimdev/lspsaga.nvim/pull/1065#issuecomment-1546541027
Hey, @glepnir, any thoughts on this one? :)

lua/lspsaga/callhierarchy.lua
```diff
@@ -122,11 +122,25 @@ function ch:call_hierarchy(item, parent)
...
         local expand_collapse = '  ' .. ui.expand
         local icon = kind[target.kind][2]
-        local className = target.uri:match('.+/(.+)[.]class[?]')
-        className = className and className or ''
+
+        -- Variable can be used to append name of class to the respective method in the outgoing preview popup.
+        local classNamePart = ''
+
+        -- Class name resolution for Java
+        if vim.bo[api.nvim_get_current_buf()].filetype == 'java' then
```
glepnir, line 130:
```suggestion
        if vim.bo[self.main_buf].filetype == 'java' then
```

glepnir: a littile busy recently sry for late

blz777: No problem :) Thank you very much!

PR #1068, opened 2023-05-13
style: function symbol missing space

protiumx:
There was a missing space for the `function` symbol.
It could make sense to wrap them all into a function that adds the extra space or any pre/suffix

fix


glepnir:
could you change the function symbol to `fx` symbol  in nerdfonts. you can find it on nerdfonts website. I forget change it on main branch .

PR #1072, opened 2023-05-14
add outline hide_kind config

daog1:
During the process of using it, I found that at some point, the outline displayed too many variables, which affected the reading of the code, so I added a hide_kind configuration item

The configuration method is like summer
```
outline = {
        auto_preview = false,
        win_width = 50,
        hide_kind={"Variable"},
        keys = {
          expand_or_jump = '<cr>',
        },
      },
```


glepnir:
like lua_ls ... I add a keyword check by using treesitter to check the node name .but still has many template variable symbols. in new version outline has two layout but is next version so I would like to merge this in current main.

PR #1077, opened 2023-05-18
feat: auto_resize option for outline window

kair8m:
Sometimes it's much easier to auto resize outline window to the maximum width of the rendered text in the outline window. Thus, this commit adds this feature by enhancing the configuration.
With `auto_resize` option enabled the window width is set to the maximum width of the rendered text in the outline


lua/lspsaga/init.lua
```diff
@@ -117,6 +117,7 @@ local default_config = {
     auto_preview = true,
     auto_refresh = true,
     auto_close = true,
+    auto_resize = true,
```
glepnir, line 120: not sure we should make it true in default .

kair8m: Yes, I agree. Changed to false by default 👍

doc/lspsaga.nvim.txt
```diff
@@ -564,6 +564,7 @@ Default options:
         auto_preview = true,
         auto_refresh = true,
         auto_close = true,
+        auto_resize = true,
```
glepnir, line 567: there is no need. it will auto generate by action.

kair8m: Ok, I will remove it then

PR #1080, opened 2023-05-21
fix: Allow multiple mappings to the same action

tsakirist:
This change fixes the inconsistency in the mappings that were applied inside the provided Lspsaga functions.

A single utility function `map_keys` has been created which automatically applies the action to all the mappings that have been provided by the user.


lua/lspsaga/rename.lua
```diff
@@ -31,20 +32,15 @@ end
 
 function rename:apply_action_keys()
   local modes = { 'i', 'n', 'v' }
+  local opts = { buffer = self.bufnr }
 
-  for i, mode in pairs(modes) do
-    if string.lower(config.rename.quit) ~= '<esc>' or mode == 'n' then
```
tsakirist, line 36: I'm not sure why `<esc>` was only mapped in `normal` mode without the user knowing about it. I know that this is to get out of `insert` mode for instance, but if the user has explicitly assigned the `quit` action to `<Esc>` I don't think it should be only mapped to `n` but in all modes.

glepnir: make sense. and can we wrap own keymap function based on `nvim_buf_set_keymap` . `vim.keymap.set` is a syntax suger based on that and `nvim_set_keymap` . mostly we just set buffer map . in `ref` branch i changed some keymap by using `nvim_buf_set_keymap` you can take a look at it :)  and Thanks for this pr 👍

tsakirist: I think `vim.keymap.set` provides a good abstraction over the internal implementation details of setting a keymap. But if you're mostly using `nvim_buf_set_keymap` and want to be explicit about that, then sure I can make the change!

glepnir: For lspsaga, most floating window keymaps are buffe rkeymaps. `vim.keymap.set` has some redundant wrapping and if else. It's not necessary for us so I replaced part of it in ref branch. So I'm thinking that we can simply wrap a function with `nvim_buf_set_keymap`. Replaces vim.keymap.set.

tsakirist: I made the change and now `map_keys` uses `nvim_buf_set_keymap` by default.

lua/lspsaga/utils.lua
```diff
@@ -0,0 +1,28 @@
...
+  return type(value) == 'string' and { value } or value
+end
+
+--- Creates a buffer local mapping.
+---@param buffer number
+---@param modes string|table<string>
+---@param keys string|table<string>
+---@param action string|function
+---@param opts? table
+function M.map_keys(buffer, modes, keys, action, opts)
```
glepnir, line 13: mode is better in opts.mode then  we can do `local mode = opts.mode or 'n'`  . `action` change to `rhs` .

lua/lspsaga/utils.lua
```diff
@@ -0,0 +1,28 @@
...
+  opts = opts or {}
+
+  if type(action) == 'function' then
+    opts.callback = action
+    action = ''
+  end
+
+  for _, mode in pairs(M.as_table(modes)) do
```
glepnir, line 21: for a list-like table use ipairs

glepnir:
and I am thinking about add this into ref not main branch. because in ref branch we have a utils.lua file

lua/lspsaga/utils.lua
tsakirist: If we add `mode` to `opts` then we would need to remove this field before we use it in `nvim_buf_set_keymap` since `opts.mode` is not valid.

tsakirist:
> and I am thinking about add this into ref not main branch. because in ref branch we have a utils.lua file

I think this should be in master too. We can merge the `utils.lua` file after `ref` branch.

glepnir:
okay current version I don't add unit test. in next version we have. so I would to know does it works fine ? then i can merge this . i don't have much time test this pr sry

tsakirist:
> okay current version I don't add unit test. in next version we have. so I would to know does it works fine ? then i can merge this . i don't have much time test this pr sry

Yes it works fine no problems. Now it is easy to use different keys for the same action like:

```lua
require("lspsaga").setup {
  finder = {
    max_height = 0.7,
    keys = {
      expand_or_jump = { "<CR>", "o" },
      quit = { "<Esc>", "q" },
      vsplit = "<C-v>",
      split = "<C-x>",
      tabnew = { "<C-t>" },
    },
  }
}
```

glepnir:
could you rename utils -> util

PR #1084, opened 2023-05-24
Add option to set initial mode for rename

nakatuba:
Thank you for providing such a great plugin.
I want to add an option to rename in insert mode because sometimes I want to change only a part of a variable or function name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -485,7 +485,7 @@ Default options:
     exec = "<CR>",
     mark = "x",
     confirm = "<CR>",
-    in_select = true,
+    initial_mode = "s"
   },
 ```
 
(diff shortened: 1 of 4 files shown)


glepnir:
so you can just  set `in_select` to false why change name ? it's break change

nakatuba:
Currently the only options for initial mode are select mode and normal mode. I want to start in insert mode.

glepnir:
why we not add an command option to do this  eg `:Lspaga rename mode=i/n/s` then it more useful . if you want change  all you can use `mode=s` if you want change part you can use `n or i`

nakatuba:
Yes, that would be better, because you can use it in different modes.

glepnir:
yup . this is more powerful then change in options . so I prefer we add a command argument for rename

nakatuba:
This pull request is closed because adding command options may change the behavior of other APIs. I'll appreciate it if you could implement this in the future.

glepnir:
will do in next version . create a feature request as trace.

nakatuba:
I see. I appreciate your response and work.

PR #1091, opened 2023-05-29
feat: virtual text along with word highlight for reference node

kair8m:
the displayed virtual text is usefull in a couple of cases: 
1) emphasise current line, sometimes it is a bit difficult to detect current line in the finder window 
2) keep track of current idx and overall count of text occurences inside buffer

render1685364021260


glepnir:
thanks looks great.

glepnir:
Oh i think you need clear it when you open this file buffer which  is loaded before

PR #1093, opened 2023-05-31
feat: autoclose peek definition and finder buffers on focus loss

kair8m:
close rendered windows for finder and peek_definition buffers when focus is lost (e.g. switching to another window)

before changes:

before_changes

after changes:

after_changes


kair8m:
@glepnir, could you take a look please? :)

kair8m:
> 

`BufLeave` is not valid event because the window can have the same buffer id as origin location, hence the event will not be emitted. It is necessary to close the window only if focus is lost for finder window, it has nothing to do with the buffer.

Also I fixed a bug when treesitter is not enabled for finder entries since vim api marked all the buffers as loaded, but treesitter highlighter is not started there :)

glepnir:
oh right in peek it will have same buffer . make sense. thanks for your contribute.

PR #1110, opened 2023-07-08
Fix: outline not showing up when symbols are not ready

hek14:
Some servers are very slow at startup. Outline will not show up when the symbols are not requested. Users may be confused  by `no response` of the command `:Lspsaga outline`. This commit will fix this.


glepnir:
not a good way. in this case  should check server status first.

PR #1129, opened 2023-07-10
fix: icon maybe nil on winbar

Decodetalkers:
Icon maybe nil, so it cannot simply concatenate


glepnir:
if config the `devicon` defaule option it always return an icon. but yes this patch is inneed.

PR #1138, opened 2023-07-11
fix(layout.float): wrong hl specification for default winopt

Jint-lzxy:
Border hl group of floats should be `FloatBorder` instead of `Border` :)


glepnir:
thanks for contribute.

PR #1149, opened 2023-07-12
fix(finder): restore winhl after buffer changes

Jint-lzxy:
Turns out that after calling `nvim_win_set_buf` for a window, all local options will be overwritten. Wrote a fix to restore the original winhl after `nvim_win_set_buf`.

### DIFF
--- a/lua/lspsaga/finder/init.lua
+++ b/lua/lspsaga/finder/init.lua
@@ -206,6 +206,10 @@ function fd:event()
         return
       end
       api.nvim_win_set_buf(self.rwinid, node.value.bufnr)
+      api.nvim_set_option_value('winhl', 'Normal:SagaNormal,FloatBorder:SagaBorder', {
+        scope = 'local',
+        win = self.rwinid,
+      })
       local range = node.value.selectionRange or node.value.range or node.value.targetSelectionRange
       api.nvim_win_set_cursor(self.rwinid, { range.start.line + 1, range.start.character })
       api.nvim_set_option_value('winbar', '', { scope = 'local', win = self.rwinid })


glepnir:
what's the problem is the border or normalflaot changed after buffer change in right window?

Jint-lzxy:
> what's the problem is the border or normalflaot changed after buffer change in right window?

Ys both `FloatBorder` and `NormalFloat` got changed to the global one:

info

glepnir:
oh that's bad. could your try min config ? I can't reproduce this

Jint-lzxy:
OK I'll try to figure out a minimum repro.

Jint-lzxy:
@glepnir Minimum repro: (`nvim -u init.lua -i NONE init.lua` and use `:Lspsaga finder` on any of the nodes in the file would lead to this issue)
```lua
------------ Setup special hl groups ------------
vim.api.nvim_set_hl(0, "NormalFloat", { fg = "#cdd6f4", bg = "#181825" })
vim.api.nvim_set_hl(0, "FloatBorder", { fg = "#181825", bg = "#181825" })
vim.api.nvim_set_hl(0, "SagaNormal", { bg = "#1e1e2e" })
vim.api.nvim_set_hl(0, "SagaBorder", { fg = "#89b4fa" })
-------------------------------------------------

local root = vim.fn.fnamemodify("./.repro", ":p")
for _, name in ipairs({ "config", "data", "state", "cache" }) do
	vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
	vim.fn.system({
		"git",
		"clone",
		"--filter=blob:none",
		"--single-branch",
		"https://github.com/folke/lazy.nvim.git",
		lazypath,
	})
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
	{
		"neovim/nvim-lspconfig",
		lazy = false,
		config = function()
			require('lspconfig').lua_ls.setup({})
		end,
	},
	{
		"nvimdev/lspsaga.nvim",
		lazy = true,
		event = "LspAttach",
		config = function()
			require("lspsaga").setup({
				finder = {
					silent = true,
					default = "def+ref+imp",
					layout = "float",
					keys = {
						shuttle = "[]",
						toggle_or_open = "<CR>",
						jump_to = "e",
						vsplit = "v",
						split = "s",
						tabe = "t",
						tabnew = "n",
						quit = "q",
						close = "<Esc>",
					},
				},
			})
		end,
		dependencies = { "nvim-tree/nvim-web-devicons", "nvim-treesitter/nvim-treesitter" },
	},
}
require("lazy").setup(plugins, {
	root = root .. "/plugins",
})
vim.opt.termguicolors = true
```

glepnir:
you missing `default =true` in your `nvim_set_hl`

Jint-lzxy:
Hmm but why would `default = true` be significant here? These highlights were defined before `lspsaga` was loaded, and this problem still exists after adding `default = true` to those function calls:

info

glepnir:
after add `default =true` it work as you expect ? when use `default =true` when lspaga highlight running it will use the highlight which you configed

image

Jint-lzxy:
Oh sorry I didn't explain the expected behavior clearly. The problem I encountered was `Lspsaga` didn't _use_ the highlights I defined (a.k.a, `SagaNormal` and `SagaBorder`) - it uses global fallback values (a.k.a, `NormalFloat` and `FloatBorder`) b/c the corresponding highlight wasn't being set at all (`winhl`). The following is the behavior I expect using the minimum repro, which would use `SagaNormal` and `SagaBorder` for rhs (Note the blue borders brought by `SagaBorder` and the background same as lhs brought by `SagaNormal`):

info

Not sure if the above-mentioned "problem" (https://github.com/nvimdev/lspsaga.nvim/pull/1149#issuecomment-1632295415) was designed as is or is a bug. Sorry if I misunderstood u.

glepnir:
I think i get your problem. for there two ways one is this. another is use namespace instead of `winhl`.

PR #1157, opened 2023-07-12
Fix: Correct diagnostic hl group name

AlexvZyl:
Saw this mistake while working on my colorscheme.


glepnir:
thanks for contribute :)

PR #1173, opened 2023-07-13
Fix error on LspAttach

chiefnoah:
It's not clear if this is the *correct* ID, but there's no `data` table on `args` so it crashes on `LspAttach`. This makes it go away, but I haven't validated that it doesn't break other behavior.


glepnir:
there must have data its your plugin management error

PR #1184, opened 2023-07-18
Replace deprecated health methods

MariaSolOs:
These methods [have been deprecated](https://neovim.io/doc/user/deprecated.html), and their use generates warnings when running `checkhealth`.

image


glepnir:
I remember it only work on head version? so there need a version check and alias

MariaSolOs:
> I remember it only work on head version? so there need a version check and alias

Yep, I was seeing those warnings in the nightly build.

PR #1188, opened 2023-07-18
chore: do not edit current bufnr

fecet:
This is the follow-up work for this https://github.com/nvimdev/lspsaga.nvim/pull/1074, as the latest update has rendered the previous changes ineffective.

### DIFF
--- a/lua/lspsaga/definition.lua
+++ b/lua/lspsaga/definition.lua
@@ -63,7 +63,7 @@ function def:close_all()
   api.nvim_del_augroup_by_name('SagaPeekdefinition')
 end
 
-function def:apply_maps(bufnr, current_buf)
+function def:apply_maps(bufnr)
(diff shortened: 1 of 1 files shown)


glepnir:
you mean it will close buffer ?

lua/lspsaga/definition.lua
```diff
@@ -256,8 +258,9 @@ function def:goto_definition(method)
     if vim.bo.modified and current_buffer == jump_destination then
       vim.cmd('write!')
     end
-
-    api.nvim_command('edit ' .. jump_destination)
```
glepnir, line 260: there maybe no necessary in goto_definition because it check before . see above .

fecet: You mean this? But that doesn't reduce loc

Issue #1199, opened 2023-07-21
Error detected while processing LspAttach Autocommands for "*":


emrhnpla02:
### Describe the bug

 ```
Error detected while processing LspAttach Autocommands for "*":
Error executing lua callback: ...share/nvim/lazy/lspsaga.nvim/lua/lspsaga/symbol/head.lua:25: unexpected event
stack traceback:
	[C]: in function 'nvim_create_autocmd'
	...share/nvim/lazy/lspsaga.nvim/lua/lspsaga/symbol/head.lua:25: in function 'buf_watcher'
	...share/nvim/lazy/lspsaga.nvim/lua/lspsaga/symbol/head.lua:180: in function <...share/nvim/lazy/lspsaga.nvim/lua/lspsaga/symbol/head.lua:165>
	[C]: in function 'nvim_exec_autocmds'
	...stalls/neovim/nightly/share/nvim/runtime/lua/vim/lsp.lua:1687: in function '_on_attach'
	...stalls/neovim/nightly/share/nvim/runtime/lua/vim/lsp.lua:1454: in function ''
	vim/_editor.lua: in function <vim/_editor.lua:0>
```
I get error above when I open any file that starts a lsp client

here's my config

```lua
local status_ok, lspsaga = pcall(require, "lspsaga")
if not status_ok then
  vim.notify("Lspsaga not found!")
  return
end

lspsaga.setup({
  lightbulb = {
    enable = true,
    enable_in_insert = true,
    sign = true,
    sign_priority = 40,
    virtual_text = false,
  },
  finder = {
    edit = "l",
    vsplit = "v",
    split = "s",
    quit = { "q", "<ESC>" },
  },
  code_action = {
    num_shortcut = true,
    keys = {
      quit = "<ESC>",
      exec = "<CR>",
    },
  },
  rename = {
    quit = "q",
  },
  symbol_in_winbar = {
    enable = false,
  },
  outline = {
    keys = {
      jump = "h",
      expand_collapse = "l",
      quit = "<ESC>",
    },
  },
  beacon = {
    enable = true,
    frequency = 10,
  },
  ui = {
    border = "rounded",
    kind = require("catppuccin.groups.integrations.lsp_saga").custom_kind(),
  },
})

keymap("n", "<C-o>", "<cmd>Lspsaga outline<CR>", opts)
keymap("n", "gd", "<cmd>Lspsaga goto_definition<CR>", opts)
[...]

PR #1201, opened 2023-07-22
fix: add more requirements to enable `lspsaga.symbol.head`

UrNightmaree:
related to #1199


glepnir:
if build with a old commit I think there should notify a update message for use new event.

UrNightmaree:
Maybe It's better to make config property that enable `lspsaga.symbol.head`, example: `lspsaga.setup{ enable_improved_symbol = true }`.

glepnir:
nope if you are using nightly I think you need follow up commit not pin a commit .

glepnir:
stylua lint fix.

PR #1253, opened 2023-08-22
Dev: fix jump and error, support terminal cwd

fengwk:
Description:

This pull request includes the following commits:
1. fix: fix cursor set for jdtls
2. fix: use lsp jump instead of vim edit to open jdt:// protocol buffer
3. feat: supports specifying terminal cwd
4. fix: fix error caused by invalid buf， maybe caused by defer function

Please review the changes. Thank you!

### DIFF
--- a/lua/lspsaga/callhierarchy.lua
+++ b/lua/lspsaga/callhierarchy.lua
@@ -303,7 +303,11 @@ function ch:peek_view()
         client.offset_encoding
       )
 
-      api.nvim_win_set_cursor(self.right_winid, { range.start.line + 1, col })
+      local right_bufnr = vim.api.nvim_win_get_buf(self.right_winid)
+      local total_lines = vim.api.nvim_buf_line_count(right_bufnr)
+      if range.start.line >= 0 and range.start.line < total_lines then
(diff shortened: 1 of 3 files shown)


glepnir:
LGTM buf need stylua format

fengwk:
I am not familiar with the Lua specifications. I made some adjustments according to the lint instructions. I think it should be like this?

lua/lspsaga/callhierarchy.lua
```diff
@@ -245,17 +245,23 @@ function ch:keymap()
         return
       end
       local data = self.method == get_method(2) and curnode.value.from or curnode.value.to
-      local fname = vim.uri_to_fname(data.uri)
       local start = data.selectionRange.start
-      local pos = { start.line + 1 }
       self:clean()
       local restore = win:minimal_restore()
-      vim.cmd[action](fname)
+      vim.cmd(action)
```
glepnir, line 251: could we do action with file name ? not action then use jump_to_location to open file if that function failed the file not open right ?

glepnir: fair enough .

PR #1261, opened 2023-08-25
do not pass _ as arguments in finder

defr0std:
_ is a global variable which can be assigned to. In fact, even here the global _ is assigned to:

https://github.com/defr0std/lspsaga.nvim/blob/8365520841a69111b8d7ae47fad5cbf88000885a/lua/lspsaga/finder/init.lua#L39

So the normal finder works just the first time. The second time, the _ variable can contain a bufnr, so the window is not initialized properly.

I also found other installed plugins which set the _ variable to various values such as empty string, empty table, etc.


glepnir:
thanks for catch it out !

PR #1283, opened 2023-09-07
open finder reference in the calling window

defr0std:
This is a small convenience improvement. Currently, when the finder uses a normal layout with sp_global, the reference is always open ('toggle_or_open') in the leftmost window. This is quite inconvenient when working with multiple vertical splits. Imagine that you search for reference in the right split, and then the reference is open in the left split, which has a completely different context.

When the finder is open, we memorize the current window id. When the reference is opened, we use that window id instead of 0 (which by this time is already the leftmost).

For float layout, or normal with sp_global=false, there is no change in behavior.

### DIFF
--- a/lua/lspsaga/finder/init.lua
+++ b/lua/lspsaga/finder/init.lua
@@ -31,6 +31,7 @@ end
 local ns = api.nvim_create_namespace('SagaFinder')
 
 function fd:init_layout()
+  self.callerwinid = api.nvim_get_current_win()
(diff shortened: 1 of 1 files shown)


glepnir:
sorry for late . should check layout ? looks like only useful on float right ?

defr0std:
No, this is useful only in normal layout with sp_global=true. In this case, the finder reference will be open in the window from which the finder was started. In all other use cases, there is no change in behavior.

Issue #1304, opened 2023-10-01
Enabling 'Outline' and navigating to another file causes error


Delice0:
### Describe the bug

```
00:03:51 msg_show Error executing lua callback: ...re/nvim/lazy/lspsaga.nvim/lua/lspsaga/symbol/outline.lua:318: Invalid window id: -1
stack traceback:
	[C]: in function 'nvim_win_get_height'
	...re/nvim/lazy/lspsaga.nvim/lua/lspsaga/symbol/outline.lua:318: in function 'create_preview_win'
	...re/nvim/lazy/lspsaga.nvim/lua/lspsaga/symbol/outline.lua:398: in function <...re/nvim/lazy/lspsaga.nvim/lua/lspsaga/symbol/outline.lua:383>
```

### Steps to reproduce

1. open two files e.g. two java files
2. enabled 'Outline' when inside one of the files
3. navigate to the other file

### Expected behavior

I expect it to NOT throw error when navigating to another file while 'Outline' is enabled.

### Neovim version (nvim -v)

latest

### lspsaga commit

latest

### Terminal name/version

iterm2

Delice0:
@glepnir still an issue.

Delice0:
```   Error  14:18:04 msg_show.lua_error Error detected while processing CursorMoved Autocommands for "<buffer=174>":
```


```
14:18:04 msg_show Error executing lua callback: ...re/nvim/lazy/lspsaga.nvim/lua/lspsaga/symbol/outline.lua:318: Invalid window id: -1
stack traceback:
	[C]: in function 'nvim_win_get_height'
	...re/nvim/lazy/lspsaga.nvim/lua/lspsaga/symbol/outline.lua:318: in function 'create_preview_win'
	...re/nvim/lazy/lspsaga.nvim/lua/lspsaga/symbol/outline.lua:399: in function <...re/nvim/lazy/lspsaga.nvim/lua/lspsaga/symbol/outline.lua:384>

```

glepnir:
I can't reproduce after patch.

Delice0:
@glepnir ok try this instead:

1. have two files open
2. enable outline inside one of the files
3. remove focus from outline. focus your cursor in your code instead
4. navigate to the other file
5. focus outline now

glepnir:
update and try thanks

PR #1306, opened 2023-10-05
fix outline autocommand group issue. follow up #1304

gwenyambira:
This fix address an error being thrown when reopening the outline, and is a follow up to #1304.
The autocommand group `outline` is initialised on line 18 of `outline.lua`.
The `clean_ctx()` function deletes the `outline` group by id (`nvim_del_augroup_by_id`). 
This function is called when closing the outline.
Attempting to reopen the outline causes an error to be thrown, as the autocommand group ID given has been deleted.
This fix recreates the `outline` group in the `outline_normal_win()` and `outline_in_float()` functions, allowing a valid ID to be used when the outline is refreshed.

### DIFF
--- a/lua/lspsaga/symbol/outline.lua
+++ b/lua/lspsaga/symbol/outline.lua
@@ -33,6 +33,7 @@ end
 local function outline_in_float()
   local win_width = api.nvim_win_get_width(0)
   local curbuf = api.nvim_get_current_buf()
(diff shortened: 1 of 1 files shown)


anasrar (commented):
Having same issue, test this PR by opening outline and jump to other files and had no error.

PR #1311, opened 2023-10-06
Fix typo to prevent multiple definition requests

alexandrelasouza:
I was trying to troubleshoot why the peek_definition was not working for `clojure-lsp`.
This message would've been helpful =)


lua/lspsaga/definition.lua
```diff
@@ -200,6 +200,7 @@ function def:definition_request(method, handler_T, args)
 
   lsp.buf_request(current_buf, method, params, function(_, result, context)
     if not result or #result == 0 then
+      self.pending_request = false
```
alexandrelasouza, line 203: When there were no results, it would never do another request.

PR #1357, opened 2023-11-28
refactor: add missing SagaSep highlight

protiumx:
## Description
This highlight is not defined by default which hides it from the user unless they check the source code in `lspasaga/winbar`


lua/lspsaga/highlight.lua
```diff
@@ -14,6 +14,7 @@ local function hi_define()
     SagaSpinner = { link = 'Statement' },
     SagaText = { link = 'Comment' },
     SagaSelect = { link = 'String' },
+    SagaSep = { fg = 'white', bg = 'none' },
```
protiumx, line 17: I think Neovim defaults to the `Normal` when this one was not found. maybe better default needs to be defined, please feel free to change it

PR #1368, opened 2023-12-27
Use nx mode to avoid remap keys

yehuohan:
This PR try to use 'nx' mode for feedkeys to avoid remap keys. So scroll keys won't break with user's mappings.


yehuohan:
@glepnir Seemed there's something unexpected in this PR. You had merged this PR, but ths latest code doesn't contain changes of this PR.
Is it because I deleted the fork after your merge?

PR #1398, opened 2024-02-24
fix: handle breaking change to vim.api.nvim_win_get_config (#1397)

### LINKED ISSUE #1397: nvim_win_get_config() returns numbers for col and row and not table
Hello,

In codeaction/preview.lua

```
local function create_preview_win(content, main_winid)
  local win_conf = api.nvim_win_get_config(main_winid)
```
Is not returning a table with [true] and [false] for row and col and this making this function crash.

```
nvim --version
NVIM v0.10.0-dev
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
```
I use quite a lot of plugins (woops) but here is where I reproduced it:

- lspconfig
- typescript-tools
- none-ls -> code action builtin eslint
- cssmodules_ls
- tailwindcss

This happens on `[1] Move to a new file` action

I temporary fixed this issue by detecting that row/col are not a table and then reassign it but it might not be the right way to do, just stopping the error without refactoring to much code. 

Not sure where the problem lies.
I will try to reproduce with a lighter config and share it.

scratchyone:
This PR fixes #1397, and adds support for the NVIM v0.10.0 `nvim_win_get_config()` API change.

It prioritizes the new API by making it the default and turning the return values into the v0.10.0 format if the user is running an older version of NVIM.

I've tested this locally and it fixes all the issues with code actions for me on `NVIM v0.10.0-dev-2393+g93c911e52-Homebrew` .

### DIFF
--- a/lua/lspsaga/codeaction/preview.lua
+++ b/lua/lspsaga/codeaction/preview.lua
@@ -92,15 +92,26 @@ local preview_buf, preview_winid
 ---default is under the given window
 local function create_preview_win(content, main_winid)
   local win_conf = api.nvim_win_get_config(main_winid)
+
+  -- In versions of NVIM before v0.10.0 `api.nvim_win_get_config` will return a table for .col and .row
+  -- see https://github.com/neovim/neovim/issues/24430
+  if (type(win_conf.col) == "table") then
+    win_conf.col = win_conf.col[false]
+  end
+
+  if (type(win_conf.row) == "table") then
+    win_conf.row = win_conf.row[false]
+  end
+
   local max_height
   local opt = {
     relative = win_conf.relative,
     win = win_conf.win,
-    col = win_conf.col[false],
+    col = win_conf.col,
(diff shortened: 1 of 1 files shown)


glepnir:
there has util is_ten as condtion i do some check in layout you can check it.

scratchyone:
> there has util is_ten as condtion i do some check in layout you can check it.

I noticed that helper function while I was reviewing the rest of the codebase for any other possible v0.10.0-related bugs. However, I was worried about future-proofness because it only checks the minor version. That approach wouldn't work properly if/when NVIM goes to version 1.0, which is why I decided to check the data-type instead. However, if you'd rather me update the commit to use the `is_ten` function I'd be happy to do that.

glepnir:
in nightly the return type lua convert is fixed so when 0.10 released i decide remove support for 0.9* so we can just use nvim_win_get_config().col/row directly.

PR #1414, opened 2024-03-30
refactor(codeaction): border margin handling + CodeActionCursorLine highlight

dmun:
I changed the hard coded margin between preview and code actions to make it 0 whenever `ui.border == 'none'`.

I also made a `code_action.cursorline` option defaulting to `true`. This feels like a more sane default since it's more of a 'completion-like' UI.

Let me know if you think it's okay, I will then update docs and other things if required!

## Before without borders

https://github.com/nvimdev/lspsaga.nvim/assets/39461169/63bc31f7-1a94-4e80-8654-a9e4f3a45000

## After without borders

https://github.com/nvimdev/lspsaga.nvim/assets/39461169/465ea5c4-927e-4eda-af84-85d6f1bb8249

## After with borders (and default highlights)

https://github.com/nvimdev/lspsaga.nvim/assets/39461169/7460e32c-83d2-4e70-af49-e48a1b48fef6

### DIFF
--- a/lua/lspsaga/codeaction/init.lua
+++ b/lua/lspsaga/codeaction/init.lua
@@ -94,8 +94,6 @@ function act:action_callback(tuples, enriched_ctx)
   -- initial position in code action window
   api.nvim_win_set_cursor(self.action_winid, { 1, 1 })
 
-  vim.opt.winhl:append('CursorLine:SagaCursorLine')
-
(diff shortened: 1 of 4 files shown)


glepnir:
maybe can give curosrline a new  highlght it's better than `CursorLine` i guess.   I will try add it in next verison. for now looks good. thanks

PR #1415, opened 2024-03-31
feat: show diagnostic with fallbacks

LintaoAmons:


lua/lspsaga/diagnostic/show.lua
```diff
@@ -336,13 +336,31 @@ function sd:show(opt)
   end)
 end
 
+local function get_diagnostic_entrys(opt)
+  if opt.fallback then
+    local scopes = { 'cursor', 'line', 'buffer' }
+    for _, scope in ipairs(scopes) do
+      local result = diag:get_diagnostic({ [scope] = true })
+      if not (next(result) == nil) then
```
glepnir, line 344: for list-like table can be simply use hash symbol `#` to check it's empty

PR #1420, opened 2024-04-08
feat: add padding+hl to code_action servers

KevinSilvester:
* Added padding before the showing server name for better readability
* Created a new highlight group for server names

---

**Before**
image

**After**
image

---

**Before**
image

**After**
image

### DIFF
--- a/lua/lspsaga/codeaction/init.lua
+++ b/lua/lspsaga/codeaction/init.lua
@@ -124,7 +124,7 @@ function act:action_callback(tuples, enriched_ctx)
   for i = 1, #content, 1 do
     local row = i - 1
     local col = content[i]:find('%]')
-    local server_start = content[i]:find('%(', max_len) - 1
+    local server_start = (content[i]:find('%(', max_len) or 1) - 1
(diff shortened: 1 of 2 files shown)


glepnir:
seems like top border is hidden ?

KevinSilvester:
I think that may be due to a separate issue with the preview window.

I haven't really looked into the cause of the issue.
But from my experience the preview window sometimes appears underneath Code Actions window when:
   * There are a lot actions available for that line
   * The cursor is nearer the top of the buffer.

Plus all the screenshots under 'Before' were using the current main branch of `lspsaga` without any changes to its code.
The is another screen recording of the window overlap when using the main branch.

https://github.com/nvimdev/lspsaga.nvim/assets/56918431/cd591e4d-f504-4bec-8e3b-057bd6452420

glepnir:
After new refactor so do we still this now ?

KevinSilvester:
Yup all good to go.
The refactor seems to have fixed the overlapping issue

image

glepnir:
oh next step is refactor the ui of code action :) diagnostic jump more fancy now :p

PR #1424, opened 2024-04-19
Change rename so that curser in insert mode at the end of variable name

kkrime:
Quality of life improvement;

I feel like this is more intuitive, as when a rename is made, more often than not, it's not a complete rename, it's just a partial change to the name.

If you don't agree with this change, would you be open to implementing this as an option the user can set instead?

### DIFF
--- a/lua/lspsaga/rename/init.lua
+++ b/lua/lspsaga/rename/init.lua
@@ -122,9 +122,10 @@ function rename:lsp_rename(args)
 
   if mode == 'i' then
     vim.cmd.startinsert()
+    vim.cmd([[normal! $]])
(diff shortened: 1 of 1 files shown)


kkrime:
Ignore, I managed to achieve this via;

```
vim.keymap.set("n", "<leader>rn", function()                                                                                                                                                                                                                                                               
vim.api.nvim_command([[:Lspsaga rename]])                                                                                                                                                                                                                                                                
local keys = vim.api.nvim_replace_termcodes('<ESC>A', true, false, true)                                                                                                                                                                                                                       
vim.api.nvim_feedkeys(keys, 'm', false)                                                                                                                                                                                                                                                                  
end, { silent = true })     
```

Issue #1417, opened 2024-04-02
Indefinite '[lspsaga] a hover request has already been sent, please wait.'


DanielVolchek:
### Describe the bug

Sometimes when using hover_doc, the pending status will get stuck and will never reset. The only way to get hover_doc working in that case is to fully restart nvim. 

https://github.com/nvimdev/lspsaga.nvim/assets/25828026/0bb0aa57-9cda-4907-ace4-86dc1e2f4bfd

### Steps to reproduce

Easiest way I've found is to go into a typescript file and try to hover_doc a keyword like return. From then on hover_doc will be stuck

### Expected behavior

hover_doc should reset if its stuck in pending

### Neovim version (nvim -v)

NVIM v0.9.5

### lspsaga commit

a4d4428

### Terminal name/version

iterm2 3.4.19

DanielVolchek:
bump, this is still not working

MyyPo:
@DanielVolchek feel free to check out this PR #1437

PR #1437, opened 2024-05-11
fix(hover): cancel prev lsp hover request on new requests instead of waiting addresses #1417

MyyPo:
Now instead of waiting for previous request to complete do a cleanup by calling the cancel function returned by `vim.lsp.buf_request` and proceed to create a new lsp request. Seems to help with both regular hover windows and `++keep` windows. 

Long-term tho it is probably better to use `vim.lsp.buf_request_all` instead of `vim.lsp.buf_request` as by [this nvim issue](https://github.com/neovim/neovim/pull/19738).

### DIFF
--- a/lua/lspsaga/hover.lua
+++ b/lua/lspsaga/hover.lua
@@ -250,7 +250,7 @@ function hover:do_request(args)
       else
         value = result.contents.value
       end
-    elseif vim.islist(result.contents) then -- MarkedString[]
+    elseif vim.tbl_islist(result.contents) then -- MarkedString[]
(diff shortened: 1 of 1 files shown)


lua/lspsaga/hover.lua
```diff
@@ -249,7 +250,7 @@ function hover:do_request(args)
       else
         value = result.contents.value
       end
-    elseif vim.tbl_islist(result.contents) then -- MarkedString[]
+    elseif vim.islist(result.contents) then -- MarkedString[]
```
glepnir, line 253: need a alias for neovim version islist is in 0.10 for 0.9 still need tbl_islist.

PR #1457, opened 2024-06-02
Fix typo and Create Luadoc

VoxelPrismatic:
1. Fixed a typo (`symbol_in_winbar.dely` --> `delay`)
2. Created luadoc so the LSP can validate your options
   - These docs are created from the latest `init.lua`, but using the descriptions from the website
image
image
image


glepnir:
thanks very nice.

PR #1482, opened 2024-08-15
feat: adds the ability to set a prefix to winbar

adriancmiranda:
This update introduces a new feature that allows users to set a custom prefix for the window bar (winbar). 

## Key Changes:

**New Configuration:** A new configuration option has been added to set a prefix that will be displayed before the current content of the winbar.

**Flexible Format:** The prefix can be a simple string and is consistently applied across all windows where the winbar is active.

## Configuration Example:

```lua
return {
  ...
  ui = {
    winbar_prefix = ' ',
    ...
  },
  ...
}
```

### Appearance

image

> before

---

image

> after

## Benefits:

**Customization:** Provides an easy way to customize the appearance of the winbar according to user preferences.


lua/lspsaga/init.lua
```diff
@@ -4,6 +4,7 @@ saga.saga_augroup = api.nvim_create_augroup('Lspsaga', { clear = true })
 
 local default_config = {
   ui = {
+    winbar_prefix = ' ',
```
glepnir, line 7: default value should be empty `''`

adriancmiranda: I will amend the commit to set the default value to an empty string (''). Thank you for pointing that out! I forgot to change it.

PR #1513, opened 2025-01-14
Nvim11 position encoding fix

LarssonMartin1998:
Make sure to always pass a offset_encoding to `make_given_range_params` and `make_position_params`. This is now a warning in nvim 11 even though the functionality hasnt changed, so it's backwards compatible.

Also slight refactorting, add util function `get_offset_encoding` with table for filtering how to get it to match new nvim api. This is used everywhere now.


glepnir:
lint fix thanks

PR #1536, opened 2025-04-07
removes hint bulb tip

zoefiri:
this was causing a weird issue where my gutter would rapidly disappear and reappear when moving my cursor, so I removed it since I don't have much use for it anyways.


zoefiri:
oops.

PR #1538, opened 2025-04-17
fix(treewide): remove deprecated uses of LSP client methods

SimonYde:
I'm not entirely sure about the expected version compatibility of this plugin, but from latest stable version the use of methods on Lsp.Client (e.g. client.supports_method) is deprecated and should be used as e.g. `client:supports_method`.

### DIFF
--- a/lua/lspsaga/callhierarchy.lua
+++ b/lua/lspsaga/callhierarchy.lua
@@ -361,7 +361,7 @@ end
 
 function ch:call_hierarchy(item, client, timer_close, curlnum)
   self.pending_request = true
-  client.request(self.method, { item = item }, function(_, res)
+  client:request(self.method, { item = item }, function(_, res)
     self.pending_request = false
     curlnum = curlnum or 0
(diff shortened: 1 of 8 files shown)


SunJun8:
Does this modification have any requirements for the version of neovim?

SunJun8:
It is recommended to revert to the previous commit before resolving compatibility issues.

glepnir:
revert need a version check some people still use 0.10x

SunJun8:
> revert need a version check some people still use 0.10x

You can also note the current Neovim version requirements in the readme.

SunJun8:
In `doc/lspsaga.nvim.txt`, there are still fields such as `For Nvim 0.8.0`

SimonYde:
> I'm not entirely sure about the expected version compatibility of this plugin, ...

I wrote this PR well aware that it might break previous compatibility. I was therefore surprised to see it merged so quickly. I'm willing to have another go at this, but I'd like to know which versions are sought to be supported. @glepnir

Issue #1336, opened 2023-10-22
Lspsaga finder def+ref show only definitions


yannick1974:
### Describe the bug

Since commit b48b807 `Lspsaga finder def+ref` shows only the definition while `Lspsaga finder ref` correctly shows the references. Reverting the commit (including the typos associated commits) makes the finder  behaves correctly — well at least as before.

I experience this problem on Python code with pyright or pylsp as LSP servers.

### Steps to reproduce

With a Python file like using default Lspsage configuration:

```python
def test():
    print("foo")


if __name__ == "__main__":
    test()
```

Invoking  `Lspsaga finder def+ref` on the `test()` shows only the definition while `Lspsaga finder ref` shows the two references, including the definition.

### Expected behavior

`Lspsaga finder def+ref` should show both the definition and the references.

### Neovim version (nvim -v)

v0.10.0-dev bc850ba

### lspsaga commit

28051f1

### Terminal name/version

konsole

glepnir:
In most languages, definitions same as Declaration. But some other languages such as C python are ignored. I think adding result filtering is feasible. Elements that are the same as range should be deleted.

PR #1550, opened 2025-08-29
Add finder option ref_opt

wiserfz:
Hi, thank you for such amazing plugin. 

I noticed the command `lspsaga finder def+ref` will display the definition and the references, but the definition also include in the references section:

### bashls

image

### rust-analyzer

image

and I also see the issue #1336 , so I open this PR and add an option `finder.ref_opt` which indicated the parameters `includeDeclaration` for the LSP RCP method `textDocument/references`, the default value is `true` and also can be  a `table` which can setup value for each LSP client for example:

```lua
local saga = require("lspsaga")

saga.setup({
    finder = {
      ref_opt = {
        bashls = false,
        rust_analyzer = false,
        lua_ls = true,
      },
    },
})
```

### bashls

image

### rust-analyzer

image

### lua_ls

image

Any feedback is appreciated.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,12 +53,14 @@ jobs:
           version: nightly
 
       - name: luajit
-        uses: leafo/gh-actions-lua@v10
+        uses: leafo/gh-actions-lua@v11
         with:
           luaVersion: "luajit-2.1"
 
       - name: luarocks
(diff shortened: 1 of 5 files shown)


wiserfz:
@glepnir hi, how about this feature? Hope it can be merged.

glepnir:
LGTM

PR #1554, opened 2025-09-27
Migrate deprectated nvim method calls

benjiwolff:
These old usages cause warnings for plugin users. I just migrated the usages that I received warnings for. There might be more.
https://neovim.io/doc/user/deprecated.html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+# Generated files
+doc/tags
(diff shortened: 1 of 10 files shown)


mTvare6:
Seems like the repo is unmaintained now. Is there a community recognised alternative fork, or any discussion of it?

